PR #865, opened 2021-03-26
PLAT-127115: Fixed Input label when value is 0

daniel-stoian-lgp:
### Checklist

- The issue was discovered in the qa-sampler for a picker with add/remove items which contains an input, with default value 0 Click on the index input field. --> Input '0'. --> Index Button's inner text is still 'Index'

### Resolution
- Updated the computed value for ButtonLabel inside InputBase to take a value of `0` as a valid value

### Additional Considerations

### Links
PLAT-127115

### Comments

Enact-DCO-1.0-Signed-off-by: Daniel Stoian daniel.stoian@lgepartner.com

### DIFF
--- a/Input/Input.js
+++ b/Input/Input.js
@@ -466,7 +466,7 @@ const InputBase = kind({
 
 	computed: {
 		buttonAriaLabel: ({placeholder, type, value}) => {
-			if (value) {
+			if (value || value === 0) {
 				type = isPasswordType(type) ? 'password' : type;
 				return calcAriaLabel('', type, type === 'number' ? value.split('') : value);
 			}
@@ -474,7 +474,11 @@ const InputBase = kind({
 			return calcAriaLabel('', null, placeholder);
 		},
 		buttonLabel: ({placeholder, type, value}) => {
-			return (isPasswordType(type) ? convertToPasswordFormat(value) : value) || typeof value === 'number' ? value.toString() : placeholder;
+			if (value || value === 0) {
+				return isPasswordType(type) ? convertToPasswordFormat(value) : value.toString();
+			} else {
+				return placeholder;
+			}
 		}
 	},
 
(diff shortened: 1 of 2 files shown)


adrian-cocoara-lgp (approved):
LGTM

Input/Input.js
```diff
@@ -474,7 +474,7 @@ const InputBase = kind({
 			return calcAriaLabel('', null, placeholder);
 		},
 		buttonLabel: ({placeholder, type, value}) => {
-			return (isPasswordType(type) ? convertToPasswordFormat(value) : value) || placeholder;
+			return (isPasswordType(type) ? convertToPasswordFormat(value) : value) || typeof value === 'number' ? value.toString() : placeholder;
```
0x64, line 477: To make this logic more readable, what about updating this code like the below pseudo code?

```
if (value's type is not one of String or Number) {
	return placeholder;
} else {
	return isPasswordType(type) ? convertToPasswordFormat(value) : value
}
```

0x64: Also, I found the same issue of value 0 during review from the upper function `buttonAriaLabel`.

```
		buttonAriaLabel: ({placeholder, type, value}) => {
			if (value) { // This line is to check whether value is null or undefined due to a missing prop value. So, type checking looks better here.
				...
			}
			...
		},
```

Input/Input.js
```diff
@@ -466,15 +466,19 @@ const InputBase = kind({
...
 				type = isPasswordType(type) ? 'password' : type;
 				return calcAriaLabel('', type, type === 'number' ? value.split('') : value);
 			}
 
 			return calcAriaLabel('', null, placeholder);
 		},
 		buttonLabel: ({placeholder, type, value}) => {
-			return (isPasswordType(type) ? convertToPasswordFormat(value) : value) || placeholder;
+			if (value || value === 0) {
+				return isPasswordType(type) ? convertToPasswordFormat(value) : value.toString();
```
0x64, line 478: It looks that we don't need `.toString()` anymore.

0x64: Oh, I see this is required. 👍

PR #888, opened 2021-04-21
PLAT-141390: Changed back key behavior on PopupTabLayout

MikyungKim:

### Checklist

When the back key pressed from the content which is the first panel, it should not close the popup but it should move the focus to tabs area.

### Resolution
Added `onKeyUp` handler to TabLayout and see if the target is from the content that is the first panel, move the spotlight to the tabs and call `stopPropagation` to not close the Popup.

### Additional Considerations

### Links
PLAT-141390

### DIFF
--- a/TabLayout/TabLayout.js
+++ b/TabLayout/TabLayout.js
@@ -275,7 +275,7 @@ const TabLayoutBase = kind({
 			const {collapsed, 'data-spotlight-id': spotlightId, type} = props;
 			const popupPanelRef = document.querySelector(`[data-spotlight-id='${spotlightId}'] .${popupTabLayoutComponentCss.panel}`);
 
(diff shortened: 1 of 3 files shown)


jeonghee27 (approved):
LGTM

TabLayout/TabLayout.js
```diff
@@ -268,6 +270,19 @@ const TabLayoutBase = kind({
 				}
 			}
 		},
+		onKeyUp: (ev, props) => {
+			const {keyCode, target} = ev;
+			const {collapsed, 'data-spotlight-id': spotlightId, type} = props;
+			const popupPanelRef = document.querySelector(`[data-spotlight-id='${spotlightId}'] .${popupTabLayoutComponentCss.panel}`);
+
+			if (forward('onKeyUp', ev, props) && type === 'popup' && is('cancel')(keyCode) && popupPanelRef.contains(target) && popupPanelRef.dataset.index === '0') {
```
0x64, line 278: Please move `forward` call out of `if` clause as it returns always true.

PR #889, opened 2021-04-21
PLAT-126939: added separate labels for navigation buttons in WizardPanels

0x64:

### Checklist

GUI requires to show button labels not as a part of a button but as a separate label along with an icon button.

### Resolution
Added `nextButtonLabel` and `prevButtonLabel` props to `WizardPanels.panel`

### Additional Considerations

### Links
PLAT-126939

### DIFF
--- a/WizardPanels/WizardPanels.js
+++ b/WizardPanels/WizardPanels.js
@@ -28,6 +28,11 @@ import css from './WizardPanels.module.less';
 const WizardPanelsContext = createContext(null);
 const DecoratedPanelBase = FloatingLayerIdProvider(PanelBase);
 const HeaderContainer = SpotlightContainerDecorator(Header);
+const ButtonLabel = (props) => (
+	<div className={css.buttonLabel}>
+		<div className={css.label} {...props} />
+	</div>
+);
 
 /**
  * A WizardPanels that has steps with corresponding panels.
@@ -122,6 +127,8 @@ const WizardPanelsBase = kind({
 		 */
 		nextButton: PropTypes.oneOfType([PropTypes.bool, EnactPropTypes.componentOverride]),
 
+		nextButtonLabel: PropTypes.string,
+
 		/**
 		 * Specifies when and how to show `nextButton` on WizardPanel.
 		 *
@@ -215,6 +222,8 @@ const WizardPanelsBase = kind({
 		 * @private
 		 */
 		prevButton: PropTypes.oneOfType([PropTypes.bool, EnactPropTypes.componentOverride]),
+
+		prevButtonLabel: PropTypes.string,
 
 		/**
 		 * Specifies when and how to show `prevButton` on WizardPanel.
@@ -348,6 +357,7 @@ const WizardPanelsBase = kind({
 		footer,
 		index,
 		nextButton,
+		nextButtonLabel,
 		nextButtonVisibility,
 		noAnimation,
 		onNextClick,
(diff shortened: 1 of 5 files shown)


samples/sampler/stories/default/WizardPanels.js
```diff
@@ -39,9 +39,14 @@ export const _WizardPanels = () => (
...
-					<Button icon="closex" aria-label="exit">
-						Exit
-					</Button>
+					<Button icon="closex" aria-label="exit" />
+				) : (
+					void 0
+				)
+			}
+			prevButtonLabel={
+				boolean('custom first Panel prevButton', Config) ? (
```
jeonghee27, line 48: Shouldn't it be written in string form?
boolean('prevButtonLabel', Config)

I am not sure how to use this prop by looking at this sample.

WizardPanels/WizardPanels.js
```diff
@@ -114,14 +119,22 @@ const WizardPanelsBase = kind({
...
 		 */
 		nextButton: PropTypes.oneOfType([PropTypes.bool, EnactPropTypes.componentOverride]),
 
+		/**
+		 * The label for the next button.
+		 *
+		 * @type {String}
+		 * @private
+		 */
+		nextButtonLabel: PropTypes.string,
```
jeonghee27, line 136: Did you try to make unit test for this?

WizardPanels/WizardPanels.js
```diff
@@ -28,6 +28,11 @@ import css from './WizardPanels.module.less';
 const WizardPanelsContext = createContext(null);
 const DecoratedPanelBase = FloatingLayerIdProvider(PanelBase);
 const HeaderContainer = SpotlightContainerDecorator(Header);
+const ButtonLabel = (props) => (
+	<div className={css.buttonLabel}>
```
jeonghee27, line 32: I don't know the GUI guide, but do we need these 2 div? 3 div are added for each button.

0x64:
@jeonghee27 Could you let me know your opinion for a general direction first? I'm not sure this approach is good to go or not. I tried to modify `Button` but it was not so successful. Also, button components are used for various container type components so I want to affect only `WizardPanels` code only.

jeonghee27:
@0x64 I like current direction 
@MikyungKim how about you?

samples/sampler/stories/default/WizardPanels.js
0x64: Well, putting new props for both labels would be better, maybe?

WizardPanels/WizardPanels.js
0x64: I didn't. 🤔  I will add if new APIs look good to reviewers.

WizardPanels/WizardPanels.js
0x64: The outer one is to make the label inline and the inner one to make the text vertically centered. If there is a better way, please let me know.

WizardPanels/WizardPanels.js
0x64: APIs are agreed, so I added unit tests.

samples/sampler/stories/default/WizardPanels.js
```diff
@@ -1,15 +1,17 @@
 import {mergeComponentMetadata} from '@enact/storybook-utils';
```
jeonghee27, line 1: Could you fix qa-sampler too?
image

jeonghee27 (approved):
LGTM except for qa-sampler

PR #894, opened 2021-04-27
PLAT-141464: Fix Panels.Header to remeasure marquee metrics when `slotSize` changed

MikyungKim:

### Checklist

In some tricky cases, by timing issue, the title of Panels.Header in FixedPopupPanel doesn't start marquee when the title is centered.

### Resolution
The reason why the marquee not starting is, when Header is centered and the `slotSize` updated, the Marquee itself thinks the props are not changed so it does not recalculate its metrics.
To be specific, when Header is centered, it calculates `slotSize` to give the size for `slotAfter` and `slotBefore` by `Measurable`. The title is in between `slotAfter` and `slotBefore` so if the size of the slots changed, the width of the title is also changing.
But the problem is, Header itself rerenders well but `Heading` component used for title and subtitle will never know they are updated or not since their prop has not changed at all. In this case, we should let them know by `Measureable` or something else.
Luckily, Marquee has `invalidateProps` for recalculating its metrics when the given prop has been changed.
So I've added `slotSize` prop to `invalidateProps` so Heading can recalculate its metrics.

### Additional Considerations
Unfortunately, it's hard to reproduce the problem with samples so I can't create test case for this.

### Links
PLAT-141464

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ The following is a curated list of changes in the Enact sandstone module, newest
 - `sandstone/FormCheckboxItem` to show correct color for `slotBefore` icon in disabled state when focused
 - `sandstone/ImageItem` to resize the image properly
 - `sandstone/Input` button label when default value is `0`
+- `sandstone/Panels.Header` to remeasure marquee metrics when `slotSize` prop changed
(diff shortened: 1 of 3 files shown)


Sangwook-Lee (approved):
LGTM!

PR #906, opened 2021-05-13
PLAT-123715: Sampler: Fix global knob setting resets when component changes

YunhoLee0:
### Checklist

The global knob setting stays even if the story changes.

### Resolution
The storybook offers object 'globals' in the decorator and story.
I added properties that need to be worked globally to the object.
(Referred link: https://storybook.js.org/docs/react/essentials/toolbars-and-globals#globals)

### Links
PLAT-123715

### DIFF
--- a/samples/sampler/src/ThemeEnvironment/ThemeEnvironment.js
+++ b/samples/sampler/src/ThemeEnvironment/ThemeEnvironment.js
@@ -133,6 +133,8 @@ const StorybookDecorator = (story, config = {}) => {
 	// Executing `story` here allows the story knobs to register and render before the global knobs below.
 	const sample = story();
 
+	const {globals} = config;
+
 	const Config = {
 		defaultProps: {
 			locale: 'en-US',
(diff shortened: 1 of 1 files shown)


samples/sampler/src/ThemeEnvironment/ThemeEnvironment.js
```diff
@@ -167,18 +169,24 @@ const StorybookDecorator = (story, config = {}) => {
 		classes.debug = true;
 	}
 
+	globals.locale = select('locale', locales, Config, globals.locale);
```
MikyungKim, line 172: What's for the 4th param..?

MikyungKim (approved):
💯 Works like :magic:! :sogood: Let's apply this to all sampler.

samples/sampler/src/ThemeEnvironment/ThemeEnvironment.js
YunhoLee0: It is the selected value. So when we change the story, a global locale is applied.

PR #910, opened 2021-05-17
PLAT-142955: Fix error when `value` prop of `Input` is 0

0x64:
### Checklist

The error occurs when `value` prop of `Input` is 0 due to `split` function call for number type value.

### Resolution
The logic is to manipulate string value, so `toString` is added to handle it properly.

### Additional Considerations

### Links
PLAT-142955


0x64:
Note: changelog should be added AFTER the current releasing process ends to avoid a conflict of changelogs.

jeonghee27 (approved):
LGTM

jeonghee27:
we omitted changelog because it occurred on specific sample.

PR #938, opened 2021-05-25
PLAT-140745: Refine logic to handle the last input type

0x64:

### Checklist

The current API to handle the last input type imported by callback prop for `SpotlightRootDecorator` and it will be better to be imported in a normal way.

### Resolution
Updated to import key APIs from the module, not through callback function prop.
Also, updated TabLayout to use the API without querying a specific node to get the type info.

### Additional Considerations

### Links
PLAT-140745

### DIFF
--- a/TabLayout/TabLayout.js
+++ b/TabLayout/TabLayout.js
@@ -26,6 +26,8 @@ import PropTypes from 'prop-types';
 import compose from 'ramda/src/compose';
 import {createContext, Fragment} from 'react';
 
+import {getLastInputType} from '../ThemeDecorator';
+
 import RefocusDecorator, {getNavigableFilter, getTabsSpotlightId} from './RefocusDecorator';
 import TabGroup from './TabGroup';
 import Tab from './Tab';
@@ -38,8 +40,7 @@ const TabLayoutContext = createContext(null);
 const TouchableCell = Touchable(Cell);
 
 const isTouchMode = () => {
-	const rootContainer = document.querySelector('#root > div');
-	return rootContainer && rootContainer.classList.contains('spotlight-input-touch');
+	return getLastInputType() === 'touch';
 };
 
 /**
--- a/TabLayout/TabLayout.js
+++ b/TabLayout/TabLayout.js
@@ -26,7 +26,7 @@ import PropTypes from 'prop-types';
 import compose from 'ramda/src/compose';
 import {createContext, Fragment} from 'react';
 
-import {getLastInputType} from '../ThemeDecorator';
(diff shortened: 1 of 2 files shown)


TabLayout/TabLayout.js
```diff
@@ -38,8 +40,7 @@ const TabLayoutContext = createContext(null);
 const TouchableCell = Touchable(Cell);
 
 const isTouchMode = () => {
-	const rootContainer = document.querySelector('#root > div');
-	return rootContainer && rootContainer.classList.contains('spotlight-input-touch');
+	return getLastInputType() === 'touch';
```
jeonghee27, line 43: Please remove "{ return "

ThemeDecorator/ThemeDecorator.js
```diff
@@ -296,4 +284,4 @@ const ThemeDecorator = hoc(defaultConfig, (config, Wrapped) => {
 });
 
 export default ThemeDecorator;
-export {ThemeDecorator};
+export {ThemeDecorator, getInputType as getLastInputType};
```
jeonghee27, line 287: I'm not sure we should change the name

jeonghee27: Please rename when import

jeonghee27 (approved):
LGTM

ThemeDecorator/ThemeDecorator.js
```diff
@@ -14,7 +14,7 @@ import {Component} from 'react';
 import classNames from 'classnames';
 import {ResolutionDecorator} from '@enact/ui/resolution';
 import {FloatingLayerDecorator} from '@enact/ui/FloatingLayer';
-import SpotlightRootDecorator from '@enact/spotlight/SpotlightRootDecorator';
+import SpotlightRootDecorator, {activateInputType, getInputType, setInputType} from '@enact/spotlight/SpotlightRootDecorator';
```
MikyungKim, line 17: I expected import `getInputType` as `getLastInputType` here.

ThemeDecorator/ThemeDecorator.js
```diff
@@ -177,7 +177,7 @@ const ThemeDecorator = hoc(defaultConfig, (config, Wrapped) => {
 		[css.bg]: !overlay
 	});
 
-	let spotlightInputType = {};
+	const spotlightInputType = {request: null};
```
MikyungKim, line 180: How about renaming this as below?
```suggestion
	const requestInputType = null;
```

MikyungKim (approved):
:lgtm:

PR #948, opened 2021-05-27
PLAT-127486: Fix Popup to have proper focus when opening with `noAnimation` is `true`

MikyungKim:

### Checklist

The original issue happened in FixedPopupPanels. When FixedPopupPanels opened with `noAnimation` prop is `true`, the last focused element got focused instead of the first element which is the result of `noAnimation` prop is `false`.

### Resolution
The difference between the two cases came from `data-spotlight-container-disabled` attribute.
When `noAnimation` prop is `false`, Popup pauses spotlight and so the container of the popup turns to be disabled. But when it's `true`, there is no such step since the node will be destroyed after all.
Spotlight stores the last focus information before the node got destroyed so the last focused element stored since the container is not disabled.
I made the container disabled when `noAnimation` prop is `true` in `getDerivedStateFromProps` phase.

### Additional Considerations
The container node will be destroyed anyway and re-create when it's open again so I don't think it has side-effects.

### Links
PLAT-127486

### DIFF
--- a/samples/sampler/stories/qa/FixedPopupPanels.js
+++ b/samples/sampler/stories/qa/FixedPopupPanels.js
@@ -314,7 +314,7 @@ export const WithVariousItems = () => {
 							<Button size="small" disabled onClick={nextPanel} onKeyDown={handleKeyDown}>Button1</Button>
 							<br />
 							<br />
-							<Button size="small">Button2</Button>
+							<Button size="small" onClick={handleClose}>Button2</Button>
 							<Button size="small" onClick={nextPanel} onKeyDown={handleKeyDown}>Button3</Button>
 							<br />
 							<br />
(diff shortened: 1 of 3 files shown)


samples/sampler/stories/qa/FixedPopupPanels.js
```diff
@@ -314,7 +314,7 @@ export const WithVariousItems = () => {
 							<Button size="small" disabled onClick={nextPanel} onKeyDown={handleKeyDown}>Button1</Button>
 							<br />
 							<br />
-							<Button size="small">Button2</Button>
+							<Button size="small" onClick={handleClose}>Button2</Button>
```
jeonghee27, line 317: How about change label to Close
```suggestion
							<Button size="small" onClick={handleClose}>Close</Button>
```

jeonghee27 (approved):
LGTM

jeonghee27 (approved):
LGTM

PR #969, opened 2021-06-22
PLAT-143663: Fixed Picker value to not marquee when changing title

MikyungKim:

### Checklist

The value of Picker marquees when adding or changing the title.

### Resolution
Marquee for Picker is under control by MarqueeController but the Heading which is used for title starts marquee on "render" so the value starts marquee on "render" also.
So I changed Heading's `marqueeOn` prop to `hover`.

### Additional Considerations

### Links
PLAT-143663


seunghoh (approved):
LGTM

PR #973, opened 2021-06-28
PLAT-145128: Fix VirtualList snapToCenter issues

MikyungKim:
### Checklist

- Sometimes, 5way down/up doesn't work.
- If the user wheels down and up repeatedly, the focused effect stays at some items.
- Sometimes, scrolls to jump by 5way without animation.

### Resolution
- Sometimes, 5way down/up doesn't work. : Reset internal flag that is critical to 5way behavior.
- If the user wheels down and up repeatedly, the focused effect stays at some items. : Added and cleared logic related scaled effect
- Sometimes, scrolls to jump by 5way without animation. : Always on `animateOnFocus` option when `snapToCenter`

### Additional Considerations

### Links
PLAT-145128

### DIFF
--- a/useScroll/useEvent.js
+++ b/useScroll/useEvent.js
@@ -113,9 +113,6 @@ const useEventFocus = (props, instances) => {
 				spottable.current.animateOnFocus = false;
 			}
 		}
-		if (themeScrollContentHandle.current.resetSnapToCenterStatus) {
-			themeScrollContentHandle.current.resetSnapToCenterStatus();
-		}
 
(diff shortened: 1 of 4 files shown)


0x64 (approved):

useScroll/useEvent.js
```diff
@@ -51,7 +49,7 @@ const useEventFocus = (props, instances) => {
 					scrollContainerHandle.current.start({
 						targetX: left,
 						targetY: top,
-						animate: spottable.current.animateOnFocus,
+						animate: spottable.current.animateOnFocus || snapToCenter,
```
jeonghee27, line 52: I think it should be resolved on PLAT-125066
Let;s revisit.

PR #975, opened 2021-06-29
PLAT-143483: Fixed Wizardpanel to not re-render the previous panel

jeonghee27:
### Checklist

WizardPanels: Items contained in the panel being removed are rendered twice

### Resolution
3 renderings are performed when the Wizard Panel is loaded.

- The 2nd render is called by useReverseTransition so I modified it to use Ref.
- The 3rd render is by `WizardPanels.Panel`, which is difficult to fix due to the current structure. The reported problem occurs in the 3rd render situation, because the key value is updated before the panel contents are updated.
So I fixed the key value is specified on `WizardPanels.Panel` instead of `WizardPanels`

### Additional Considerations

The current WizardPanel.panel is returning null and updates the parent state. 
This structure needs long-term fix.

### Links
PLAT-143483

### Comments


0x64 (approved):

MikyungKim (approved):
:lgtm:

PR #994, opened 2021-07-08
Revert "PLAT-143483: Fixed Wizardpanel to not re-render the previous panel"

jeonghee27:
Reverts enactjs/sandstone#975


jeonghee27:
close in favor of #995

PR #1044, opened 2021-07-22
WRN-1380: Migrate Dropdown unit tests to Testing Library

stanca-pop-lgp:

### Checklist

Migrated unit tests from Enzyme to Testing Library for Dropdown component.

### Resolution

### Additional Considerations

### Links
WRN-1380


Dropdown/tests/Dropdown-specs.js
```diff
@@ -1,261 +1,271 @@
...
+			<Dropdown title={title}>
+				{updatedChildren}
+			</Dropdown>
+		);
 
 		const expected = 5;
-		const actual = dropDown.children().length;
+		const actual = updatedChildren.length;
 
 		expect(actual).toBe(expected);
```
alexandrumorariu, line 185: Here you can use toHaveLength() instead, but it is up to you.

Dropdown/tests/Dropdown-specs.js
```diff
@@ -1,261 +1,271 @@
...
+
+		rerender(
+			<Dropdown title={title}>
+				{updatedChildren}
+			</Dropdown>
+		);
 
 		const expected = 5;
-		const actual = dropDown.children().length;
+		const actual = updatedChildren.length;
```
daniel-stoian-lgp, line 183: Unfortunately, this does not verify that the actual dropdown has 5 options, but just that an array is updated. You will proably need to change the test to render an opened dropdown and check the number of children of a node(something like screen.getByRole("list").children.length

stanca-pop-lgp: You are right. Thank you for the suggestion!

alexandrumorariu (approved):
LGTM

daniel-stoian-lgp (approved):

PR #1048, opened 2021-07-26
WRN-3709: Block wheel event in VideoPlayer

changgilee:
### Checklist

When scrolling wheel in video player, media list should be show up.
But in current status, try to seek media file.

### Resolution
Wheel event handler in Slider is accept event more higher priority than video panel.
For resolve this issue, new prop is added to slider and used in MediaSlider.

### Additional Considerations
N/A

### Links
WRN-3709

### Comments
N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,6 @@ The following is a curated list of changes in the Enact sandstone module, newest
 
 - `sandstone/Slider` prop `noWheelEvent` to disable wheel event handler
 
-
 ## [2.0.0-rc.5] - 2021-07-22
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 The following is a curated list of changes in the Enact sandstone module, newest changes on the top.
 
+## [unreleased]
+
+### Added
+
+- `sandstone/Slider` prop `noWheelEvent` to disable wheel event handler
+
+
 ## [2.0.0-rc.5] - 2021-07-22
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The following is a curated list of changes in the Enact sandstone module, newest
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


0x64:
I updated the checklist in the description; this PR does not break API but adds a single prop.

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact sandstone module, newest changes on the top.
 
+## [unreleased]
+
+### Added
+
+- `sandstone/Slider` prop `noWheelEvent` to disable wheel event handler
```
0x64, line 9: What about `noChangeByWheel` or simply `noWheel`? `noWheelEvent` indicates _an implementation manner_, and it looks better to name a prop to mean _a certain hehavior by some user input_.

MediaPlayer/MediaSlider.js
```diff
@@ -85,6 +85,7 @@ const MediaSliderBase = kind({
 			<div className={className}>
 				<Slider
 					{...rest}
+					noWheelEvent
```
0x64, line 88: Please make it alphabetically ordered.

Slider/Slider.js
```diff
@@ -106,6 +106,7 @@ const SliderBase = (props) => {
 	const nativeEventHandlers = useHandlers({
 		onWheel: handle(
 			forProp('disabled', false),
+			forProp('noWheelEvent', false),
```
0x64, line 109: Please check wheel input on a normal slider also.
Since there is no default value for `noWheelEvent`, `forProp` will return `false` in normal sliders. We don't recommend setting a falsy default value, so what about fixing this like below?

```suggestion
			not(forProp('noWheelEvent', true)),
```

0x64 (changes requested):
There is a prop bleeding in `render` function. Please delete a newly added prop before pass props to render.

CHANGELOG.md
seunghoh: I prefer `noWheel`

0x64 (approved):
:lgtm:

seunghoh (approved):
LGTM

PR #1051, opened 2021-07-27
WRN-1349: Migrated unit tests for Icon and DateTimeDecorator to testing-library. Removed unit tests for DebounceDecorator

daniel-stoian-lgp:
### Checklist

Migrated unit tests for Icon and DateTimeDecorator to testing-library. 
Removed unit tests for DebounceDecorator because they couldn't be migrated to testing-library

### Resolution

### Additional Considerations

### Links
WRN-1349

### Comments

Enact-DCO-1.0-Signed-off-by: Daniel Stoian daniel.stoian@lgepartner.com


internal/DateTime/tests/DateTimeDecorator-specs.js
```diff
@@ -1,31 +1,34 @@
...
-				<Picker
-					title="Date"
-					value={new Date(2000, 0, 1, 12, 30)}
-					locale="en-US"
-				/>
-			);
+	test('should accept an updated JavaScript Date for its value prop', () => {
+		const Picker = DateTimeDecorator({}, function PickerBase ({locale, title, value}) {
+			const minuteValue = value.getMinutes();
+			return <div locale={locale} title={title} >{minuteValue}</div>;
```
daniel-stoian-lgp, line 10: The prop is available for reading only if it displayed in the DOM. That's why the view is modified

internal/DateTime/tests/DateTimeDecorator-specs.js
```diff
@@ -1,31 +1,34 @@
...
+			/>
+		);
 
-			expect(actual).toBe(expected);
-		}
+		const expected = '45';
+		const actual = screen.getByTitle('Date').textContent;
+		expect(actual).toBe(expected);
+	}
 	);
```
alexandrumorariu, line 33: Place both these parenthesis on the same line.

daniel-stoian-lgp: done

alexandrumorariu (approved):
LGTM

internal/DateTime/tests/DateTimeDecorator-specs.js
```diff
@@ -1,31 +1,33 @@
...
-				/>
-			);
-
-			subject.setProps({
-				value: new Date(2000, 0, 1, 12, 45)
-			});
+	test('should accept an updated JavaScript Date for its value prop', () => {
+		const Picker = DateTimeDecorator({}, function PickerBase ({locale, title, value}) {
+			const minuteValue = value.getMinutes();
+			return <div locale={locale} title={title} >{minuteValue}</div>;
```
stanca-pop-lgp, line 10: Please remove the empty space before >

internal/DateTime/tests/DateTimeDecorator-specs.js
```diff
@@ -1,31 +1,33 @@
...
+			<Picker
+				title="Date"
+				value={new Date(2000, 0, 1, 12, 45)}
+				locale="en-US"
+			/>
+		);
 
+		const expected = '45';
+		const actual = screen.getByTitle('Date').textContent;
+		expect(actual).toBe(expected);
```
stanca-pop-lgp, line 31: Please add an empty line before expect.

stanca-pop-lgp (approved):
LGTM

Icon/tests/Icon-specs.js
```diff
@@ -1,159 +1,107 @@
...
+		expect(icon).toHaveClass(expected);
+	});
+
+	test('should support arbitrary custom numeric sizes', () => {
+		render(<Icon data-testid="icon" size={96}>star</Icon>);
+
+		const expected = '--icon-size';
+		const icon = screen.getByTestId('icon').style;
+
+		expect(icon).toHaveProperty('0', expected);
```
MikyungKim, line 94: What is this `0` for?

internal/DebounceDecorator/tests/DebounceDecorator-specs.js
```diff
@@ -1,101 +0,0 @@
-import {mount} from 'enzyme';
-import DebounceDecorator from '../';
-
-describe('DebounceDecorator', () => {
-	test(
-		'should emit the event after the delay',
```
MikyungKim, line 6: I was able to make a test for `DebounceDecorator` after some searching. :)
Well, there are two things we should have in mind to write this kind of test.
First, we need to fire custom events, in this case, `onChange` and `onCancel`. For firing custom events with React Testing Library, we have `createEvent` method.
Secondly, we need to use `useRef` to attach the handler for these custom events to dispatch. Unlike the DOM events, we should use `addEventListener` and `remove EventListener`.
In conclusion, the problem here was not the DOM is not rendered with `render` method, but it needs some special handling for custom events.
Let me show you one example(the first one) and assign the rest to you.
```
test(
	'should emit the event after the delay',
	(done) => {
		const Component = DebounceDecorator(
			{debounce: 'onChange', delay: 100},
			function Base ({onChange}) {
				const ref = useRef(null);

				useEffect(() => {
					const {current} = ref;
					current.addEventListener('onChange', onChange);

					return () => {
						current.removeEventListener('onChange', onChange);
					};
				}, [onChange]);

				return <div ref={ref}>Test</div>;
			}
		);

		const spy = jest.fn();

		render(<Component onChange={spy} />);

		const elem = screen.getByText('Test');
		fireEvent(elem, createEvent('onChange', elem));

		expect(spy).not.toHaveBeenCalled();

		setTimeout(() => {
			expect(spy).toHaveBeenCalled();
			done();
		}, 150);
	}
);
```

Icon/tests/Icon-specs.js
daniel-stoian-lgp: well, the existing test is supposed to show that an '--icon-size' property exists, in comparison with the next test, where the value is '--icon-size' exists.
Since the function toHaveStyle() checks if an element has specific css properties with specific values applied,  I couldn't use it here, so I had to check if '--icon-size' exists:

image

internal/DebounceDecorator/tests/DebounceDecorator-specs.js
daniel-stoian-lgp: Thank you!

Icon/tests/Icon-specs.js
MikyungKim: 🤔 I agree to remove this test case since the next one covers it.

MikyungKim (approved):
LGTM

PR #1054, opened 2021-08-03
WRN-3886: Item: Fix to update marquee when `slotAfter` or `slotBefore` changed

MikyungKim:

### Checklist

When changing `slotAfter` or `slotBefore` dynamically, sometimes marquee doesn't update properly.

### Resolution
The reason why this issue happens is that we don't re-calculate the distance for marquee after `slotAfter` or `slotBefore` changed.
I've added `MarqueeDecorator` to add `invalidateProps` to trigger the re-calculation of the distance for those props.

### Additional Considerations
When we implement the Item for the first time, in #5, we've changed `MarqueeDecorator` to `MarqueeController` but we didn't care about `invalidateProps` which was there for `inline`. I've also fixed it.

### Links
WRN-3886

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact sandstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `sandstone/Item` to marquee properly when `slotAfter` or `slotBefore` changed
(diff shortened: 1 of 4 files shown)


jeonghee27 (approved):
LGTM

0x64 (approved):

PR #1055, opened 2021-08-04
WRN-3773: Fix 5-way navigation of list items when item's internal content is updated

0x64:

### Checklist

5-way navigation to an *updated (re-rendered)* item fails when item's content is updated without updating a list.

### Resolution
The issue occurs due to an invalid DOM node info which is used to target 5-way navigation.
`enact/ui`'s logic is updated in enactjs/enact/pull/2937 to provide safe item container nodes rather than item nodes which could be invalidated by item-internal rendering, and this PR updates `VirtualList` to use the item container nodes.

### Additional Considerations

### Links
WRN-3773
enactjs/enact/pull/2937

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact sandstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


0x64:
Close as we decided not to apply this change.

PR #1056, opened 2021-08-06
WRN-3886: Fix SwitchItem restarts marquee when clicked

MikyungKim:
### Checklist

SwitchItem which uses Item inside restarts marquee when clicked

### Resolution
Since `Slottable` gives slots as new objects, if we compare the reference of slots, it will change whenever it renders. This lead to the re-start of the marquee.
Instead of this, we need to compare the size of slots so we can avoid this situation.
To notice the size of the node changed, I used `useMeasurable` and keep the refs regardless of `slotAfter` or `slotBefore` prop existence.
Also, `shrink` prop never gets down to `Marquee` since `Cell` is consuming, I added `inline` prop for inavlidateProps.

### Additional Considerations
I've changed the DOM structure of the Item, maybe it can cause some regression.

### Links
WRN-3886

### DIFF
--- a/Item/Item.js
+++ b/Item/Item.js
@@ -14,11 +14,11 @@ import classnames from 'classnames';
 import EnactPropTypes from '@enact/core/internal/prop-types';
 import kind from '@enact/core/kind';
 import Spottable from '@enact/spotlight/Spottable';
-import Pure from '@enact/ui/internal/Pure';
+import Slottable from '@enact/ui/Slottable';
 import {ItemBase as UiItemBase, ItemDecorator as UiItemDecorator} from '@enact/ui/Item';
 import {Cell, Layout, Row} from '@enact/ui/Layout';
 import {useMeasurable} from '@enact/ui/Measurable';
-import Slottable from '@enact/ui/Slottable';
+import Pure from '@enact/ui/internal/Pure';
 import compose from 'ramda/src/compose';
 import PropTypes from 'prop-types';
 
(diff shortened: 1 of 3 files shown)


Item/Item.js
```diff
@@ -240,31 +280,48 @@ const ItemBase = kind({
...
+					</span>
+				</Cell>
 			</UiItemBase>
 		);
 	}
 });
 
+const ItemMeasurementDecorator = (Wrapped) => {
+	return function ItemMeasurementDecorator (props) { // eslint-disable-line no-shadow
+		const {ref: slotAfterRef, measurement: {width: slotAfterWidth = 0} = {}} = useMeasurable() || {};
```
0x64, line 307:
```suggestion
		const {ref: slotAfterRef, measurement: {width: slotAfterWidth = 0} = {}} = useMeasurable();
```

Item/Item.js
```diff
@@ -240,31 +280,48 @@ const ItemBase = kind({
...
+		const {ref: slotBeforeRef, measurement: {width: slotBeforeWidth = 0} = {}} = useMeasurable() || {};
```
0x64, line 308:
```suggestion
		const {ref: slotBeforeRef, measurement: {width: slotBeforeWidth = 0} = {}} = useMeasurable();
```

Item/Item.js
MikyungKim: ✔️

Item/Item.js
MikyungKim: ✔️

0x64 (approved):

PR #1065, opened 2021-08-20
WRN-5464: Fixed Scroller to move focus via up/down keys from scroll thumb when the content is short but the scrollbar is visible

MikyungKim:

### Checklist

When the Scroller has short content and the scrollbar is visible, the focus is not moving via up down key from the scroll thumb. 

### Resolution
When the content is short, the `--scrollbar-thumb-progress-ratio` is NaN since the client size and scrollable size is the same.
We didn't care about this case when we move the focus from scroll thumb. I've added a guard to care about this case.

### Additional Considerations

### Links
WRN-5464

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact sandstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


0x64 (approved):
💯

seunghoh (approved):
LGTM

PR #1067, opened 2021-08-24
WRN-5513: TimePicker: Fixed abnormal minute animation in some locales

MikyungKim:

### Checklist

* [ ] I have read and understand the [contribution guide](http://enactjs.com/docs/developer-guide/contributing/)

Minute picker's animation looks weird in some locales(e.g. ko-KR, ar-SA,...)

### Resolution
We display pickers in Timepicker with flex and in some locales, the start position would have decimal points since we determine the width of the pickers by the number of '0'.
It seems like an issue with blink but we couldn't make a pure HTML example that reproduces the issue.
Meanwhile, we've found that if we make the picker's width integer value, the animation seems ok.
So I've added min-width to sizingPlaceholder.

### Additional Considerations

### Links
WRN-5513

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact sandstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `sandstone/TimePicker` abnormal minute animation in some locales
+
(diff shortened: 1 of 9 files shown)


0x64 (approved):
:lgtm:

PR #1070, opened 2021-08-25
WRN-2730: QA-Sampler: Fix the title is the same as the story name

YunhoLee0:
### Checklist

Before migrating storybook 6.x, the title of each sample was the same as its story name.
There seems to be a bug when migrating the storybook.

### Resolution
Make the title of each sample the same as the story name.

### Links
WRN-2730


samples/sampler/src/ThemeEnvironment/ThemeEnvironment.js
```diff
@@ -155,6 +155,8 @@ const StorybookDecorator = (story, config = {}) => {
 		groupId: 'Development'
 	};
 
+	const componentName = config.kind.replace(/^([^/]+)\//, '');
```
juwonjeong, line 158: I think we can also use regExp in this way.
```suggestion
	const componentName = config.kind.replace(/^\w+\//, '');
```

juwonjeong:
LGTM

juwonjeong (approved):
LGTM

samples/sampler/src/ThemeEnvironment/ThemeEnvironment.js
```diff
@@ -178,7 +180,7 @@ const StorybookDecorator = (story, config = {}) => {
 	return (
 		<Theme
 			className={classnames(classes)}
-			title={`${config.kind}`.replace(/\//g, ' ').trim()}
+			title={componentName === config.name ? `Sandstone ${componentName}` : `${componentName} ${config.name}`}
```
MikyungKim, line 183: How about
```suggestion
			title={componentName === config.name ? `${config.kind}`.replace(/\//g, ' ').trim() : `${componentName} ${config.name}`}
```
for other theme usage?

YunhoLee0: I think that's better than before. Thank you!
It's fixed.

MikyungKim (commented):
Would you fix this in agate too?

MikyungKim (approved):
👍 Now we can see the right QA sampler names again. Thanks.

PR #1073, opened 2021-09-01
WRN-5909: Sync up UI-Test and TC.

dongsuwon:
Enact-DCO-1.0-Signed-off-by: Dongsu Won (dongsu.won@lgepartner.com)

### Checklist

- Need to change the TC number in the UI-Test

### Resolution
- Change the TC number in the UI-Test

### Additional Considerations

### Links
[//]:# (Related issues, references)
WRN-5909

### Comments
As the TC moves from Jira to CLM change the TC number in the UI-Test.


tests/ui/specs/SwitchItem/SwitchItem-specs.js
```diff
@@ -23,12 +23,12 @@ describe.skip('SwitchItem', function () {
 		});
 
 		describe('5-way', function () {
-			it('should select the item when selected - [GT-21377]', function () {
+			it('should select the item when selected', function () {
```
juwonjeong, line 26: Would it be possible to match TC GT-21377 with TC QWT-2527? 
```suggestion
			it('should select the item when selected - [QWT-2527]', function () {
```

tests/ui/specs/TabLayout/VerticalTabsWithIcons/VerticalTabsWithIcons-specs.js
```diff
@@ -11,8 +11,7 @@ describe('TabLayout', function () {
 		describe('collapsing/expanding behavior', function () {
 			describe('5-way interaction', function () {
 
-				// this covers GT-28257 step 3
-				it('should collapse tabs when focus is moved to a Spottable component in the content container', function () {
+				it('should collapse tabs when focus is moved to a Spottable component in the content container - [QWT-2758]', function () {
```
juwonjeong, line 14: Can we mention step number on line 18 and 35 instead?
```
// Step 3-2: 5-way right to Spottable component 
// Step 4: Back to the tabs
```

tests/ui/specs/TabLayout/VerticalTabsWithoutIcons/VerticalTabsWithoutIcons-specs.js
```diff
@@ -11,8 +11,7 @@ describe('TabLayout', function () {
 		describe('collapsing/expanding behavior', function () {
 			describe('5-way interaction', function () {
 
-				// this covers GT-28259 step 3
-				it('should collapse tabs when focus is moved to a Spottable component in the content container', function () {
+				it('should collapse tabs when focus is moved to a Spottable component in the content container - [QWT-2756]', function () {
```
juwonjeong, line 14: Same as VerticalTabsWithIcons-specs.js
```
// Step 3-2: 5-way right to Spottable component 
// Step 4: Back to the tabs
```

tests/ui/specs/SwitchItem/SwitchItem-specs.js
dongsuwon: As the QWT-2527 test is the pointer mode test, and this test is a 5-way test, I don't think this test and the test of QWT-2527 are the same. I think the test on 50 line is the same as QWT-2527, what do you think?

juwonjeong: Sorry, I forgot to check the test mode:( I think we can just leave it as you fixed it.

juwonjeong (approved):
LGTM

PR #1082, opened 2021-09-15
WRN-7368: VirtualList: Ignore keyEvent when the data-index is unreliable

jeonghee27:
### Checklist

VirtualList: sometimes the focus moves to the previous item while scrolling

### Resolution
I don't know why the data-index to be updated outputs in the js code. (VL code, Sportlight code..., onFocus)
However, in the keyDown event handler, the previous data-index outputs.
I presume it's caching the DOM properties inside the browser.

Workaround : Ignore key event when the data-index is unreliable

### Additional Considerations
It's not accurate because of the low reproduced rate, but it seems to be reproduced only in Chromium-based browsers.

### Links
WRN-7368

### Comments


MikyungKim (approved):
:lgtm:

PR #1101, opened 2021-11-03
WRN-6753: Fixed Cross-Platform Enact iOS: VideoPlayer - A video does not start to play

stanca-pop-lgp:

### Checklist

Added `playsInline` attribute to Video component. If set to true, the video starts playing on mobile iOS even if it is not entirely visible or on fullscreen.

### Resolution

### Additional Considerations

### Links
WRN-6753

### DIFF
--- a/VideoPlayer/Video.js
+++ b/VideoPlayer/Video.js
@@ -69,18 +69,6 @@ const VideoBase = class extends Component {
 		mediaComponent: EnactPropTypes.renderableOverride,
 
 		/**
-		 * A Boolean attribute indicating that the video is to be played "inline",
-		 * that is within the element's playback area.
-		 *
-		 * Note that the absence of this attribute does not imply that the video will always be played in fullscreen.
-		 *
-		 * @type {Boolean}
-		 * @public
-		 * @default true
-		 */
-		playsInline: PropTypes.bool,
-
-		/**
 		 * The video source to be preloaded. Expects a `<source>` node.
 		 *
 		 * @type {Node}
--- a/VideoPlayer/Video.js
+++ b/VideoPlayer/Video.js
@@ -69,12 +69,24 @@ const VideoBase = class extends Component {
 		mediaComponent: EnactPropTypes.renderableOverride,
 
 		/**
+		 * A Boolean attribute indicating that the video is to be played "inline",
+		 * that is within the element's playback area.
+		 *
+		 * Note that the absence of this attribute does not imply that the video will always be played in fullscreen.
+		 *
+		 * @type {Boolean}
+		 * @public
+		 * @default true
+		 */
+		playsInline: PropTypes.bool,
+
+		/**
 		 * The video source to be preloaded. Expects a `<source>` node.
 		 *
 		 * @type {Node}
 		 * @public
 		 */
-		preloadSource:  PropTypes.node,
+		preloadSource: PropTypes.node,
 
 		/**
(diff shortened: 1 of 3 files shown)


VideoPlayer/Video.js
```diff
@@ -253,6 +265,7 @@ const VideoBase = class extends Component {
 						controls={false}
 						key={sourceKey}
 						mediaComponent={mediaComponent}
+						playsInline
```
daniel-stoian-lgp, line 268: this is always set to true. should it be like this?

stanca-pop-lgp: Thank you for this observation. I took the liberty of not making it available to changing. So "playsInline" will always be true. I refactored accordingly.

daniel-stoian-lgp (approved):
LGTM

jeonghee27:
Sorry, I don't understand most of this issue.

Expected Result: A video plays.
Actual Result: A video does not play even if a loading animation disappears.
Is this issue about autoplay?
The `playInline` option (As I konw, it play without switching to full screen) solve this problem?
I've tested with iPhone 6s (chrome 78 and Safary), but I don't catch the difference before and after the fix, and it still doesn't play. In what environment was it verified?

jeonghee27:
Does the playsInline attribute pass to the <video> Dom? They both (Before and After) change to full scene mode and play after I touch to play. It does not play automatically, I tested it on iPhone 12.

stanca-pop-lgp:
> Sorry, I don't understand most of this issue.
> 
> Expected Result: A video plays. Actual Result: A video does not play even if a loading animation disappears.
> 
> Is this issue about autoplay? The `playInline` option (As far as I konw, it play video without switching to full screen) solve this problem? I've tested with iPhone 6s (chrome 78 and Safary), but I don't catch the difference before and after the fix, and it still doesn't play. In what environment was it verified?

@jeonghee27 I tested with iPhone 8 (Chrome v. 95.0.4638.50). Before adding the `playsInline` attribute the video was not playing after loading animation disappeared (neither autoplay, nor if I clicked play). Adding it solved the problem on my device.

stanca-pop-lgp:
> Does the playsInline attribute pass to the video DOM? They both (Before and After) change to full scene mode and play after I touch to play. It does not play automatically, I tested it on iPhone 12.

I see... well I will try testing it on iPhone 8 plus (so far this is my only extra option). Could someone on your team test it on a different iPhone? Do you think that there is not a bug or it's a different kind of problem? Thank you.

jeonghee27:
> I see... well I will try testing it on iPhone 8 plus (so far this is my only extra option). Could someone on your team test it on a different iPhone? Do you think that there is not a bug or it's a different kind of problem? Thank you.

While handling the Cross-Platform iOS issue, did you not check the operation in the actual iOS?  
I personally rented devices from iPhone users and checked them out.
It seems that the result is the same even if I borrow another device. Have you checked that props are properly passed to the Video DOM?
From what I analyzed, the "VideoPlayer" component used by the sampler is not same with the "Video" component that you fixed.

jeonghee27:
> (neither autoplay, nor if I clicked play).

That's weird. On the iPhone 12 and 13 I tested, it was played in full screen when I touch for play. But It didn't play at all on iPhone 6.

stanca-pop-lgp:
PR closed as this fix does not resolve anything and the bug cannot be reproduced on iOS 15.

PR #1109, opened 2021-11-15
WRN-6762: Cross-Platform Enact Android: qa-scroller/qa-VirtualGridList - when changing 'width', 'dataSize', VKB flashes and goes away

adrian-cocoara-lgp:
Enact-DCO-1.0-Signed-off-by: Adrian Cocoara adrian.cocoara@lgepartner.com

### Checklist

Fixed qa-scroller, qa-virtualGridList, samples on Android devices so the VKB is always displayed and you can modify the input data.

When the VKB is displaying, a page resizing is triggered, Input loses focus and the VKB disappears. Scroller and ScrollerNative have a handleResizeWindow methods defined which tells the Scroller to blur the focused item. I've made a small change to that method, if the focused item is an input, the input should always be focused.

### Resolution

### Additional Considerations

### Links

### DIFF
--- a/useScroll/useScroll.js
+++ b/useScroll/useScroll.js
@@ -235,7 +235,7 @@ const useThemeScroll = (props, instances) => {
 	function handleResizeWindow () {
 		const focusedItem = Spotlight.getCurrent();
 
-		if (focusedItem) {
+		if (focusedItem && focusedItem.nodeName !== 'INPUT') {
 			focusedItem.blur();
 		}
 	}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact sandstone module, newest changes on the top.
 
+## [unreleased]
+
+### Changed
+
+- `sandstone/useScroll` handleResizeWindow method to not blur the focused item if its type is input
+
 ## [2.0.4] - 2021-11-01
 
 ### Added


daniel-stoian-lgp (approved):
LGTM

daniel-stoian-lgp (approved):
LGTM

alexandrumorariu (approved):
LGTM

useScroll/useScroll.js
```diff
@@ -235,7 +235,7 @@ const useThemeScroll = (props, instances) => {
 	function handleResizeWindow () {
 		const focusedItem = Spotlight.getCurrent();
 
-		if (focusedItem) {
+		if (focusedItem && focusedItem.nodeName !== 'INPUT') {
```
MikyungKim, line 238: Thanks for the PR.
But this fix has still some issues in some cases.
For example, let's say there is a list that has inputs in each item. When we scroll to the middle of the list and click on the input, then the scroll position goes to the top. WIth your fix, the VKB will be still there. You can refer to the logic from https://github.com/enactjs/enact/pull/2362 which introduced this code.
Is there any event or method to notify VKB is opened so that we can prevent the issue more specifically? 

Also, have you tested on iOS devices? The issue is not reproduced on them it seems `resize` event is not firing.
Would you please check why the issue is not reproduced on iOS devices?

MikyungKim:
Closing this in favor of WRN-12916.

PR #1132, opened 2021-12-22
WRN-12845 Fixed Picker Joined not changing value on each Tap

vladut-clapou-lgp:
### Checklist

Sometimes the next value displays only after 2 taps

### Resolution
For Chrome on touch devices, the order in which the functions 'handleDown' and 'handleMouseDown', which are bind to the 'onMouseDown' event , are called in reversed order than in PC devices, thus I added an 'onTouchEvent' which is called before the 'onMouseDown' for touch devices and for that event if the platform is 'androidChrome' or 'ios' I will call the 'handleMouseDown' function.
The bug is not reproduced on IOS Safari. Adding onTouchStart to that platform would lead to a double value change on a tap.

### Additional Considerations

### Links
WRN-12845

### Comments
Enact-DCO-1.0-Signed-off-by: Clapou George Vladut vladut.clapou@lgepartner.com


adrian-cocoara-lgp (approved):
LGTM

daniel-stoian-lgp (approved):

vladut-clapou-lgp:
Closing in favor of #1148

PR #1138, opened 2021-12-30
WRN-12676: Use web font(thai) instead of system font to avoid truncating tall strings

juwonjeong:

### Checklist

On android devices, some tall strings(th-TH locale) is truncated since they use system font.
Make web font to be rendered instead of system font for this specific case.

### Resolution
@font-face rule related to thai font will be stacked onto Manual Sandstone Locale Fonts rules.

### Additional Considerations
Need to be tested with enact:feature/WRN-12676 branch (https://github.com/enactjs/enact/pull/2990)

### Links
WRN-12676


jeonghee27 (approved):
LGTM

PR #1140, opened 2022-01-03
WRN-6748: Fixed Cross-Platform Enact iOS: The placeholder of the input is biased downward

stanca-pop-lgp:

### Checklist

Adjusted line-height for small inputField so that input text is centered vertically.

### Links
WRN-6748

### DIFF
--- a/Input/InputField.module.less
+++ b/Input/InputField.module.less
@@ -101,6 +101,7 @@
 		.input {
 			font-size: @sand-inputfield-small-font-size;
 			height: @sand-inputfield-small-height;
+			line-height: @sand-inputfield-small-height;
 		}
 
 		.inputHighlight {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact sandstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `sandstone/Input.InputField` size 'small' line-height to center text vertically
+
 ## [2.1.2] - 2021-12-22
 
 - Fixed samples build issue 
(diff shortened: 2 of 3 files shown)


daniel-stoian-lgp (approved):

Input/InputField.module.less
```diff
@@ -101,6 +101,7 @@
 		.input {
 			font-size: @sand-inputfield-small-font-size;
 			height: @sand-inputfield-small-height;
+			line-height: @sand-inputfield-small-height;
```
seunghoh, line 104: Well... this change could affect tall glyph languages. 3.5 rem -> 2.75 rem is huge.
@juwonjeong Would you please test with our tall glyph? 
@stanca-pop-lgp Rather than changing size, would you please search more why vertical-align: middle doesn't work on iOS and any other workaround?

seunghoh (changes requested):

Input/InputField.module.less
stanca-pop-lgp: I tested with tall characters and it doesn't seem to be a problem, especially with the merge of #1138.
before: 
image
image
after:
image
image

seunghoh: @stanca-pop-lgp Can you answer my question above? 
>@stanca-pop-lgp Rather than changing size, would you please search more about why vertical-align: middle doesn't work on iOS and any other workaround?

stanca-pop-lgp: @seunghoh Yes, sorry, I will continue investigating. I did not find yet "vertical-align: middle" to be a problem on iOS. But will carry on with the research.

seunghoh (approved):
LGTM. Thanks for the detailed analysis and fix!

PR #1146, opened 2022-01-10
WRN-12696: CrossPlatform: Fixed overlay type Alert layout when width is narrow

MikyungKim:

### Checklist

With a narrow width, the content of the overlay type Alert is clipped. 

### Resolution
See if the width is narrower than the content of the Alert, put the buttons below.

### Additional Considerations

### Links
WRN-12696


0x64 (approved):
:lgtm:

Alert/Alert.js
```diff
@@ -167,15 +169,23 @@ const AlertBase = kind({
...
 			},
 			type
-		)
+		),
+		overflow: () => {
+			if (typeof window !== 'undefined') {
+				return window.innerWidth < ri.scale(1200); // If you will change this value, please change @sand-alert-overlay-content-width too.
+			}
+
+			return false;
```
jeonghee27, line 178: image
image
image

It is still clipped on some resolution :(
Should we add the button container width?

```suggestion
		overflow: () => {
			if (typeof window !== 'undefined') {
				const buttonContainerWidth = 0 //buttons? ri.scale(1098) : 0; // @sand-button-max-width: 900px and there are some margin.
				return window.innerWidth < ri.scale(1200) + buttonContainerWidth; // If you will change this value, please change @sand-alert-overlay-content-width too.
			}

			return false;
```

MikyungKim: Ok, I updated to use `@sand-button-min-width` value for minimizing the side effect.

PR #1190, opened 2022-03-11
Feature/wrn 12925

juwonjeong:
### Checklist

After refactoring, https://github.com/enactjs/enact/pull/2465 all language glyphs render in ther designated font regardless of the current locale.
Css rules that reduce font style differences between fonts are no longer needed.

### Resolution
Remove css rules in which the selector start with `.enact-locale-non-latin` from .less files

### Additional Considerations

### Links
WRN-12925

### DIFF
--- a/styles/mixins.less
+++ b/styles/mixins.less
@@ -122,7 +122,6 @@
 // ThemeDecorator to override the target of the rules.
 .sand-text-base(@font-size: @sand-body-font-size; @target: normal) {
 	.sand-font({
-		font-weight: @sand-non-latin-font-weight;
 	}, @target);
 
 	font-weight: normal;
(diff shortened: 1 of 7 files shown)


styles/mixins.less
```diff
@@ -263,23 +244,15 @@
 }
 
 .sand-large-button-text() {
-	.sand-font(@sand-button-font-family; @sand-non-latin-button-font-family; {
-		font-size: @sand-non-latin-button-large-font-size;
-		font-style: @sand-non-latin-button-font-style;
-		font-weight: @sand-non-latin-button-font-weight;
-	});
+	.sand-font(@sand-button-font-family; @sand-non-latin-button-font-family);
```
jeonghee27, line 247:
```suggestion
	.sand-font(@sand-button-font-family); 	// We need to discuss the scope of deletion whether we should remove non-latin related variables, or remove the provided mixin (deleting all "non-latin" related code from sandstone).

```

PR #1223, opened 2022-04-19
WRO-636: InputFieldSpotlightDecorator: Refactored to not use unnecessary setState

sjro:
### Checklist

When we run unit test from sandstone, we have a warning that says
```
console.error
Warning: An update to InputSpotlightDecorator inside a test was not wrapped in act(...).
```
It is because `InputFieldSpotlightDecorator` invokes `setState` whenever focus change.
To remove this warning, we'd like to refactor it to not use an unnecessary setState structure.

### Resolution
Refactored not to use the state because there is no need to re-render by replacing the state with internal variables.

### Additional Considerations

### Links
WRO-636

### Comments
Enact-DCO-1.0-Signed-off-by: SJ RO (sj.ro@lge.com)


juwonjeong (approved):
I think we may need rename function updateComponent(located in line 170), 
My suggestion is updateStatusOfFocus but I think it should be discussed with merger.
Except it, LGTM

Input/InputFieldSpotlightDecorator.js
```diff
@@ -141,78 +141,84 @@ const InputSpotlightDecorator = hoc(defaultConfig, (config, Wrapped) => {
...
 
 				if (!noLockPointer) {
-					releasePointer(this.state.node);
+					releasePointer(this.node);
 				}
 			}
 		}
 
-		updateFocus = (prevState) => {
+		updateComponent () {
```
jeonghee27, line 170: Can't we keep the previous interface?
updateFocus = (prevState)?

Is ther any reason that we should hav two function `updateComponent` and `updateFocus`?

Input/InputFieldSpotlightDecorator.js
```diff
@@ -141,78 +141,84 @@ const InputSpotlightDecorator = hoc(defaultConfig, (config, Wrapped) => {
...
 
 		blur = () => {
-			this.setState((state) => (
-				state.focused || state.node ? {focused: null, node: null} : null
-			));
+			if (this.focused || this.node) {
+				this.focused = null;
+				this.node = null;
+				this.updateComponent();
+			}
```
jeonghee27, line 221: else is missed.

sjro: Since the effect of setState(null) is not to call render , the else case is unnecessary.

Input/InputFieldSpotlightDecorator.js
sjro: I fixed it.

Input/InputFieldSpotlightDecorator.js
```diff
@@ -141,78 +141,84 @@ const InputSpotlightDecorator = hoc(defaultConfig, (config, Wrapped) => {
...
 
 				if (!noLockPointer) {
-					releasePointer(this.state.node);
+					releasePointer(this.node);
 				}
 			}
 		}
 
-		updateFocus = (prevState) => {
+		updateFocus = (prevFocusStatus) => {
```
jeonghee27, line 172:
```suggestion
		updateFocus = () => {
```
I think the pam is unnecessary

Input/InputFieldSpotlightDecorator.js
```diff
@@ -141,78 +141,84 @@ const InputSpotlightDecorator = hoc(defaultConfig, (config, Wrapped) => {
...
+			this.fromMouse = false;
 			this.ariaHidden = props.noReadoutOnFocus || null;
 			this.paused = new Pause('InputSpotlightDecorator');
 			this.handleKeyDown = handleKeyDown.bind(this);
-		}
-
-		componentDidUpdate (_, prevState) {
-			this.ariaHidden = null;
-			this.updateFocus(prevState);
+			this.prevFocusStatus = {
```
jeonghee27, line 150:
```suggestion
			this.prevStatus = {
```

Input/InputFieldSpotlightDecorator.js
```diff
@@ -141,78 +141,84 @@ const InputSpotlightDecorator = hoc(defaultConfig, (config, Wrapped) => {
...
+			prevFocusStatus.focused = this.focused;
+			prevFocusStatus.node = this.node;
 		};
 
 		focus = (focused, node, fromMouse) => {
-			this.setState({focused, node, fromMouse});
+				this.focused = focused;
+				this.node = node;
+				this.fromMouse = fromMouse;
+				this.updateFocus(this.prevFocusStatus);
```
MikyungKim, line 213: Check the indentation.

jeonghee27 (approved):
LGTM

MikyungKim (approved):
:shipit:

Input/InputFieldSpotlightDecorator.js
```diff
@@ -230,7 +236,7 @@ const InputSpotlightDecorator = hoc(defaultConfig, (config, Wrapped) => {
 						this.blur();
 						forwardBlur(ev, this.props);
 					} else {
-						this.focusDecorator(ev.currentTarget);
+						this.fromMouse = false;
```
seunghoh, line 239: @sjro Let's also set the `ev.currentTarget` to the node since we are not sure if there are any edge cases.

PR #1233, opened 2022-05-06
WRO-709: Fixed EditableWrapper to calculate items position properly in RTL locales

juwonjeong:

### Checklist

Added support for RTL behavior in editable wrapper.

### Resolution
Fixed `EditableWrapper` to calculate the offset and direction of items in reverse in the rtl locale.

### Additional Considerations

### Links
WRO-709


Scroller/EditableWrapper.js
```diff
@@ -248,20 +249,23 @@ const EditableWrapper = (props) => {
...
 
 	const moveItemsByKeyDown = useCallback((ev) => {
 		const {keyCode} = ev;
 		const container = scrollContentRef.current;
 		const {itemWidth, prevToIndex} = mutableRef.current;
-		const toIndex = is('left', keyCode) ? prevToIndex - 1 : prevToIndex + 1;
+		const {rtl} = scrollContainerHandle.current;
 
-		const itemLeft = toIndex * itemWidth - container.scrollLeft;
+		const toIndex = (rtl ^ is('left', keyCode)) ? prevToIndex - 1 : prevToIndex + 1;
```
0x64, line 260: `typeof rtl` could be 'boolean' or 'undefined', so it looks that `rtl != is('left', keyCode)` is also possible, right? If bitwise XOR operator is preferred, a kind of type casting hack is recommended for safety like `!rtl ^ !is('left', keycode)`.
There is a related Q&A on [stackoverflow](https://stackoverflow.com/questions/4540422/why-is-there-no-logical-xor-in-javascript).

Scroller/EditableWrapper.js
```diff
@@ -311,10 +320,9 @@ const EditableWrapper = (props) => {
 			reset();
 
 			if (lastInputType === 'scroll') {
-				const offset = itemWidth * (lastMouseClientX > scrollContentCenter ? 1 : -1);
-
+				const offset = itemWidth * (rtl ^ (lastMouseClientX > scrollContentCenter) ? 1 : -1);
 				scrollContainerHandle.current.start({
-					targetX: scrollContentNode.scrollLeft + offset,
+					targetX: scrollContentNode.scrollLeft * (rtl ? -1 : 1) + offset,
```
0x64, line 325: > targetX: (scrollContentNode.scrollLeft + itemWidth * (lastMouseClientX > scrollContentCenter) ? 1 : -1) * (rtl ? -1 : 1),

We have another way if we merge `offset` calculation into this line.

0x64:
When there is no item, the following run-time error occurs;
> TypeError: Cannot read properties of undefined (reading 'nextElementSibling')

0x64 (approved):
:lgtm:

Scroller/EditableWrapper.js
```diff
@@ -311,10 +320,8 @@ const EditableWrapper = (props) => {
 			reset();
 
 			if (lastInputType === 'scroll') {
-				const offset = itemWidth * (lastMouseClientX > scrollContentCenter ? 1 : -1);
-
 				scrollContainerHandle.current.start({
-					targetX: scrollContentNode.scrollLeft + offset,
+					targetX: (scrollContentNode.scrollLeft + itemWidth * (lastMouseClientX > scrollContentCenter ? 1 : -1)) * (rtl ? -1 : 1),
```
jeonghee27, line 324: How about using a format similar to the below?

const offset = itemWidth * (!rtl ^ !(lastMouseClientX > scrollContentCenter) ? 1 : - 1);

Scroller/EditableWrapper.js
```diff
@@ -364,14 +371,16 @@ const EditableWrapper = (props) => {
...
-		if (!mutableRef.current.itemWidth) {
-			const item = wrapperRef.current?.children[0];
+		const scrollContentNode = scrollContentRef.current;
+		const item = scrollContainerHandle.current.rtl ? wrapperRef.current?.children[dataSize - 1] : wrapperRef.current?.children[0];
+		if (item) {
 			const neighbor = item.nextElementSibling || item.previousElementSibling;
+
 			mutableRef.current.itemWidth = Math.abs(item.offsetLeft - neighbor?.offsetLeft);
-			mutableRef.current.centeredOffset = centered ? item.getBoundingClientRect().x : 0;
+			mutableRef.current.centeredOffset = (centered && (scrollContentNode.getBoundingClientRect().width > mutableRef.current.itemWidth * dataSize)) ? item.getBoundingClientRect().x : 0;
```
jeonghee27, line 380: Why we need this new condition? please leave the reason.

Scroller/EditableWrapper.js
```diff
@@ -400,13 +409,15 @@ const EditableWrapper = (props) => {
 		// addEventListener to moveItems while scrolled
 		const scrollContentNode = scrollContentRef.current;
 		const scrollContentCenter = scrollContentNode.getBoundingClientRect().width / 2;
+		const scrollContentRight = scrollContentNode.getBoundingClientRect().right;
```
jeonghee27, line 412: getBoundingClientRect() need DOM accessing. and it takes time..
Let's reduce the DOM accessing.

Or I think we can store the scrollContentNode like  `mutableRef.current.centeredOffset`.
 `mutableRef.current.scrollContentRect?`.



```suggestion
		const scrollContentRect = scrollContentNode.getBoundingClientRect();
		...
```

Scroller/EditableWrapper.js
juwonjeong: In LTR, centeredOffset set with the value of mutableRef.current.child[0].getBoudningClientRect().x 
But in RTL, centeredOffset cannot be figured out with the value of getBoundingClientRect().x as shown in the picture below.
20220518_152733_centered_offset
We need the value of green line to set centeredOffset instead of red line(getBoundingClientRect().x).

Scroller/EditableWrapper.js
juwonjeong: It's fixed as suggestion.

Scroller/EditableWrapper.js
juwonjeong: Yes it would be better to use format similar to the below. It's fixed.

PR #1247, opened 2022-05-17
WRO-582: [Sandstone-useScroll] Wrap the local functions in useScrollBase with useCallback

stanca-pop-lgp:
### Checklist

There are many local functions in custom Hooks of VirtualList and Scroller. When calling those custom Hooks, the functions are re-defined if those functions are not wrapped with useCallback. That would make JavaScript performance lower. By wrapping those functions with useCallback, we could prevent redefining them again if those functions's dependencies are not changed.

### Resolution
Wrapped the local functions in` useScroll` with "useCallback" am reordered them to be called after being defined.

### Additional Considerations

### Links
WRO-582

### DIFF
--- a/useScroll/useScroll.js
+++ b/useScroll/useScroll.js
@@ -17,7 +17,7 @@ import {getRect, intersects} from '@enact/spotlight/src/utils';
 import {assignPropertiesOf, constants, useScrollBase} from '@enact/ui/useScroll';
 import utilDOM from '@enact/ui/useScroll/utilDOM';
 import utilEvent from '@enact/ui/useScroll/utilEvent';
-import {useContext, useRef} from 'react';
+import {useCallback, useContext, useRef} from 'react';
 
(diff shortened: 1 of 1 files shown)


alexandrumorariu (approved):
LGTM

useScroll/useScroll.js
```diff
@@ -223,10 +223,10 @@ const useThemeScroll = (props, instances) => {
 			contextSharedState.set(ev, props);
 			contextSharedState.set(`${id}.scrollPosition`, {x, y});
 		}
-	}
+	}, [contextSharedState, props]);
```
0x64, line 226: `props` is always a new object, so `props.id` would be the right dep. It looks missing by accident. :)

0x64: Right. It could be okay in this line if we define `id` as destructed variable from `props`, but it looks unclear what impact this way may have. I'll look into this.

PR #1263, opened 2022-06-14
WRO-4784: Added A11y feature for editable scroller

jeonghee27:
### Checklist

Need to support A11y feature on for editable scroller

### Resolution
Specs : 
* When focus item in Scroller (I fixed sampler. It should be implemented on App)
<item label> + <"Edit mode to press and hold OK key")
* When entering edit mode and focusing for the first time
<item label> + <"Move left and right or press up key to delete")
*  press left / right
<item label> + <"moved to the left of"> + <item label.position+1>
<item label> + <"moved to the right of"> + <item label.position-1>
App should add `data-item-label` on the each item's container
* press left/right when focus on leftmost/rightmost
<"Leftmost">
<"Rightmost">
* (edit mode) press up  (I fixed sampler. It should be implemented on App)
<"Delete"> + <"button">
* done (OK key in edit mode)
<item label> + <"move complete">
=> Since the focus element is read again after re-rendering, it is implemented to call `annouce` after a arbitrary 300ms.
There is other way to modify the label twice in the app, But I thought this was not good way to guide.  So I just used setTimout.

### Additional Considerations

### Links
WRO-4784

### Comments

### DIFF
--- a/Scroller/EditableWrapper.js
+++ b/Scroller/EditableWrapper.js
@@ -4,9 +4,13 @@ import {is} from '@enact/core/keymap';
 import {mergeClassNameMaps} from '@enact/core/util';
 import Spotlight from '@enact/spotlight';
 import Accelerator from '@enact/spotlight/Accelerator';
+import {Announce} from '@enact/ui/AnnounceDecorator';
 import classNames from 'classnames';
+import IString from 'ilib/lib/IString';
 import PropTypes from 'prop-types';
 import {useCallback, useEffect, useRef} from 'react';
+
+import $L from '../internal/$L';
 
 import componentCss from './EditableWrapper.module.less';
 
@@ -65,6 +69,7 @@ const EditableWrapper = (props) => {
 
 		// DOM elements
 		selectedItem: null,
+		selectedItemLabel: '',
 		rearrangedItems: [],
 
 		// Indices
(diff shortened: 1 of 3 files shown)


seunghoh (commented):
hm.. I have some considerations
1. Do we really need `data-item-label`? It works as public prop but too specific cases
2. What if we just read "left/right" then, we probably no need to use `data-item-label`

One more thing is that I don't see the sample code when the item is deleted.

jeonghee27:
>hm.. I have some considerations
>1. Do we really need data-item-label? It works as public prop but too specific cases
>2. What if we just read "left/right" then, we probably no need to use data-item-label
>One more thing is that I don't see the sample code when the item is deleted.

1. I changed to use `aria-label` instead of `data-item-label`.
2. It is inevitable to receive a `label` from the outermost DOM of each item, because it is not known what structure the actual item will be composed of or whether the textContext will always exist.
3. "Item deleted" spec has been removed

seunghoh (approved):
LGTM

Scroller/EditableWrapper.js
```diff
@@ -351,6 +406,13 @@ const EditableWrapper = (props) => {
 					const orders = finalizeOrders();
 					forwardCustom('onComplete', () => ({orders}))({}, editable);
 					reset();
+
+					setTimeout(() => {
+						announceRef.current.announce(
+							selectedItemLabel + $L('move complete'),
+							true
+						);
+					}, 300); // An arbitrary delay at a level that is not ignored by the new focus element
```
0x64, line 415: Let's make a constant for this delay since it could be updated later if requested.

PR #1279, opened 2022-07-08
WRO-7341: Fixed qa-a11y panel sample to restore focus properly

juwonjeong:

### Checklist

In qa-a11y panel sample, when navigating from panel 1 to panel 0, focus did not properly restored to the last focused item.

If the `virtualList` in panel 0 does not set spotlightId, data-spotlight-id is set to container-#. And when switching to panel 0, the data-spotlight-id of panel 0 is updated, for example from container-9 to container-13.

In this case, the spotlight cannot restore focus to the `virtualList` as the spotlight remember the lastFocusedElement information as old data (e.g., container-9).

### Resolution
Added `spotlightId` prop in `VirtualList` to restore focus properly when it remounts while transitioning the panel.

### Additional Considerations
I think the virtualList's spotlightId deletion is a regression from https://github.com/enactjs/sandstone/pull/628. 

### Links
WRO-7341

### DIFF
--- a/samples/qa-a11y/src/views/Panels.js
+++ b/samples/qa-a11y/src/views/Panels.js
@@ -33,6 +33,7 @@ const PanelsView = () => {
 					dataSize={itemList.length}
 					itemRenderer={customItem}
 					itemSize={ri.scale(156)}
+					spotlightId="virtual-list"
 				/>
 			</Panel>
 			<Panel>
--- a/samples/qa-a11y/src/views/Panels.js
+++ b/samples/qa-a11y/src/views/Panels.js
@@ -31,9 +31,10 @@ const PanelsView = () => {
 				<Header title="Panel 0" />
 				<VirtualList
 					dataSize={itemList.length}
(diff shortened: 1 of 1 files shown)


samples/qa-a11y/src/views/Panels.js
```diff
@@ -33,6 +33,7 @@ const PanelsView = () => {
 					dataSize={itemList.length}
 					itemRenderer={customItem}
 					itemSize={ri.scale(156)}
+					spotlightId="virtual-list"
```
jeonghee27, line 36:
```suggestion
					dataSize={itemList.length}
					id={'virtualList_$' + index}
					itemRenderer={customItem}
					itemSize={ri.scale(156)}
					spotlightId={'virtualList_$' + index}
```

id also need to restore scroll position.

and.. I prefer keep the previous format. (it's not affect the functions. but just a preference)

jeonghee27 (approved):
LGTM

PR #1281, opened 2022-07-11
WRO-8225: Fixed scoll by hover to work properly when `scrollMode` is translate

jeonghee27:
### Checklist

hoverToScroll props is not work properly with translate scrollMode

### Resolution

Cause : 
In Translate mode, the timing when the `update()` is called by the `onScroll` event is after `RequsetAnimation`(RAF).
Therefore, it does not `stop` and the `RAF` is executed forever.

Resolution:
So, in addition to calling `cancelAnimationFrame` in the `stopRaf` function, we should use more certain way to prevent the next `RAF` execution.
In this patch, I use the `hoveredPosition` variable we already have to check if the `RAF` has stopped to prevent further `RAF`s.

FYI : The current invoking order is as follows.
`startRaf` -> `requsetAnimationFrame` -> `scrollByHover` (Call scrollTo) -> `startRaf` ->  ......

* Native Scroller 
`startRaf` -> Call `requsetAnimationFrame` id=1 -> Execute `requsetAnimationFrame` id=1 -> `scrollByHover` (Call `scrollTo`) -> `startRaf` -> Call `requsetAnimationFrame` id=2  -> `update` -> `stopRaf` -> `cancelAnimationFrame` id=2 

* Translate
`startRaf` -> Call `requsetAnimationFrame` id=1 -> Execute `requsetAnimationFrame` id=1 -> `scrollByHover` (Call `scrollTo`) -> `startRaf` -> Call `requsetAnimationFrame` id=2  -> Execute `requsetAnimationFrame` id=2 -> `scrollByHover` (Call `scrollTo`)  -> `Update` -> `StopRaf` -> `cancelAnimationFrame` id=2  -> `startRaf`  -> ....

### Additional Considerations
Another bug was found where the scrollbar was not visible when scroll by hover in translate mode.
this is because showScrollbarTrack function in the ui/useScroll refer wrong isVerticalScrollbarVisible value . 
The issue is not reproduced after page refresh(f5), it happens when user change the scrollMode dynamically on Sampler.
This seems to be related to WRO-582.
So, the issue won't be handled on this issue. Hopefully this will be resolved in WRO-582 when functions are changed to useCallback.

### Links
WRO-8225

### Comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact sandstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `sandstone/Scroller` and `sandstone/VirtualList` to scroll properly by hover when `hoverToScroll` is `true` and `scrollMode` is translate
+
 ## [2.5.0-rc.2] - 2022-07-06
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The following is a curated list of changes in the Enact sandstone module, newest
 
 ### Fixed
 
-- `sandstone/Scroller` and `sandstone/VirtualList` to scroll properly by hover when `hoverToScroll` is `true` and `scrollMode` is translate
+- `sandstone/Scroller` and `sandstone/VirtualList` to scroll properly by hover when `hoverToScroll` is `true` and `scrollMode` is `translate`
 
 ## [2.5.0-rc.2] - 2022-07-06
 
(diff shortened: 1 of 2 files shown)


0x64 (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact sandstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `sandstone/Scroller` and `sandstone/VirtualList` to scroll properly by hover when `hoverToScroll` is `true` and `scrollMode` is translate
```
```suggestion
- `sandstone/Scroller` and `sandstone/VirtualList` to scroll properly by hover when `hoverToScroll` is `true` and `scrollMode` is `translate`
```

MikyungKim (approved):
:lgtm:

PR #1289, opened 2022-07-21
WRO-679: Remove aria-hidden in Input component

jeonghee27:
### Checklist

Need refactoring to remove `aria-hidden` in Input component

### Resolution
Revmoe `noReadoutOnFocus` prop and `aria-hidden` prop in Input component.
To meet the A11y UX specifications, I  use `role="region"` and `aria-labelledby` to read title and subtile when open Input popup  instead `aria-live="alert"`.

With `aria-live="alert"`, read all children element, so we needed `aria-hidden` to avoid reading the focused element twice.
But With `role="region"` and `aria-labelledby`, We can only read the title/subtitle. and the focus item will be read after.
I apply the same method used by Panel.

* The text read when Input popup open
AS-IS : go to previous Title SubTitle placeholder string  Input field
TO-BE: Title SubTitle placeholder string

### Additional Considerations

According to the UX documentation, title is mandatory in inputPopup. So I added title and subtitle to a11y sample.

### Links
WRO-679

### Comments

### DIFF
--- a/Input/Input.js
+++ b/Input/Input.js
@@ -372,7 +372,8 @@ const InputPopupBase = kind({
 		minLength,
 		...rest
 	}) => {
-
+		const id = `inputPopup`;
+		const ariaLabelledBy = popupAriaLabel ? null : `${id}_title ${id}_subtitle`;
 		const inputProps = extractInputFieldProps(rest);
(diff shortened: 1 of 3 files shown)


seunghoh (approved):
💯 
For the Integrator, This PR also changed some title of Input sample as the title and subtitle are added

PR #1293, opened 2022-07-26
WRO-478: POC for item reordering in VirtualList

0x64:
### Checklist

POC for item reordering in VirtualList

### Resolution
The former POC implementation of #1189 is totally refined.
- Changed to wait for press-and-hold input to edit items' order
- Changed to get a single object-typed `editable` prop rather than several individual props
- Added a selected effect for a sample

### Additional Considerations
There are so many bugs and missing behaviors.
- Touch input is not supported yet.
- UI could be broken when page up/down key is pressed during reordering.
- UI could be broken or ugly when boundaries are changed.
- It is unstable when various input devices are used simultaneously.
- I cannot guarantee that it works great even with a single input device.
- Performance optimization is required.

### Links
WRO-478
WRN-17866 (#1189)

### Comments


seunghoh:
@0x64 Would you check the Travis error?

0x64:
> @0x64 Would you check the Travis error?

@seunghoh, 
Travis fails since the companion enact branch is not matched. I did not update `travis.yml` as I thought this branch would be simply a reference for a new feature that is not officially taken in yet.
For maintainers, I will update `travis.yml`. :)

PR #1314, opened 2022-09-07
WRO-9259: Fix some stories are not shown with storybook 6.5

MikyungKim:
### Checklist

After migrating Storybook 6.5, sampler- KeyGuide, qa-sampler- Dropdown - with array of children objects, and Text - Languages stories are not opened.

### Resolution
The issue is caused by `jsxDecorator` function from `@storybook/react/dist/esm/client/docs/jsxDecorator.js` (https://github.com/storybookjs/storybook/blob/main/app/react/src/client/docs/jsxDecorator.tsx#L206)
When I looked at this code, this is for the show code section in Docs tab in the sampler. So I think we could work around it by skipping the part that throws errors since it's not related to rendering the story.
We could "skip" this code by setting docs.source.code of story's parameter according to https://github.com/storybookjs/storybook/blob/main/app/react/src/client/docs/jsxDecorator.tsx#L158.
If we skip this code, I confirmed that the three stories(sampler- KeyGuide, qa-sampler- Dropdown - with array of children objects, and Text - Languages) are rendered well.

### Additional Considerations
I put a note for work around and we could remove this workaround after storybook fixes the issue(https://github.com/storybookjs/storybook/issues/18269).
Also, I fixed the preview of Docs is not showing properly by adding `inlineStories: false`.
According to https://storybook.js.org/docs/react/writing-docs/docs-page#inline-stories-vs-iframe-stories, the param became `true` by default, so our preview is not showing well.

### Links
WRO-9259

### Comments

### DIFF
--- a/samples/sampler/.qa-storybook/preview.js
+++ b/samples/sampler/.qa-storybook/preview.js
@@ -53,6 +53,7 @@ export const parameters = {
 		container: DocsContainer,
 		page: DocsPage,
 		iframeHeight: 360,
+		inlineStories: false,
 		theme: themes.light
 	},
 	options: {
--- a/samples/sampler/.storybook/preview.js
+++ b/samples/sampler/.storybook/preview.js
@@ -52,6 +52,7 @@ export const parameters = {
 		container: DocsContainer,
 		page: DocsPage,
 		iframeHeight: 360,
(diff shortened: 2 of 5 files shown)


0x64 (approved):
Good enough!

PR #1323, opened 2022-09-19
WRO-12043: Increased code coverage for Slider and Spinner

daniel-stoian-lgp:
### Checklist

Added missing tests for 
- Slider: drag and drop and for wheel events
- Spinner: default aria label

### Resolution

### Additional Considerations
Util function for drag and drop was copied for the example on the testing library website: https://testing-library.com/docs/example-drag/

### Links
WRO-12043

### Comments

Enact-DCO-1.0-Signed-off-by: Daniel Stoian [daniel.stoian@lgepartner.com](mailto:daniel.stoian@lgepartner.com)

### DIFF
--- a/Slider/tests/Slider-specs.js
+++ b/Slider/tests/Slider-specs.js
@@ -1,5 +1,5 @@
 import '@testing-library/jest-dom';
-import {fireEvent, render, screen} from '@testing-library/react';
+import {act, fireEvent, render, screen} from '@testing-library/react';
 
 import Slider from '../Slider';
 
@@ -18,11 +18,6 @@ const getElementClientCenter = (element) => {
 	return {x: left + width / 2, y: top + height / 2};
 };
 
-const sleep = ms =>
-	new Promise(resolve => {
-		setTimeout(resolve, ms);
-	});
-
 const drag = async (element, {delta, steps = 1}) => {
 	const from = getElementClientCenter(element);
 	const to = {x: from.x + delta.x, y: from.y + delta.y};
@@ -36,7 +31,7 @@ const drag = async (element, {delta, steps = 1}) => {
 	for (let i = 0; i < steps; i++) {
 		current.clientX += step.x;
 		current.clientY += step.y;
-		await sleep(1000 / steps);
+		act(() => jest.advanceTimersByTime(1000 / steps))
 		fireEvent.mouseMove(element, current);
 	}
 	fireEvent.mouseUp(element, current);
@@ -95,7 +90,7 @@ describe('Slider', () => {
 		await drag(slider, {delta: {x: 50, y: 0}});
 
 		const expectedAttribute = 'aria-valuetext';
-		const unexpectedValue = '50';
+		const unexpectedValue = '50 change a value with left right button';
 
 		expect(slider).not.toHaveAttribute(expectedAttribute, unexpectedValue);
 	});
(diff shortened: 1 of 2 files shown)


alexandrumorariu (approved):
LGTM

Slider/tests/Slider-specs.js
```diff
@@ -58,6 +87,19 @@ describe('Slider', () => {
 		expect(slider).toHaveClass(expected);
 	});
 
+	test('should change value of slider on drag', async () => {
+		render(<Slider activateOnSelect defaultValue={50} />);
+		const slider = screen.getByRole('slider');
+
+		activate(slider);
+		await drag(slider, {delta: {x: 50, y: 0}});
```
jeonghee27, line 95: Interesting!! 

But in my environment the test passes even if I remove this line.
Although aria-valuetext is changed after dragging, it returns a NaN value.
Could you check again? 

We may need to change to clicking instead of dragging.

Slider/tests/Slider-specs.js
```diff
@@ -13,6 +13,35 @@ const rightKeyDown = keyDown(39);
 const upKeyDown = keyDown(38);
 const downKeyDown = keyDown(40);
 
+const getElementClientCenter = (element) => {
+	const {left, top, width, height} = element.getBoundingClientRect();
+	return {x: left + width / 2, y: top + height / 2};
+};
+
+const sleep = ms =>
```
jeonghee27, line 21: I'm not sure this kind of code works or not.
maybe we replaced all setTimeout to "advanceTimersByTime".
in https://github.com/enactjs/sandstone/pull/1216/files

 @sjro  What do you think?

jeonghee27 (approved):

Slider/tests/Slider-specs.js
daniel-stoian-lgp: you are right. i put the wrong "unexpectedValue". the initial value is "50 change a value with left right button" , not "50"

Spinner/tests/Spinner-specs.js
```diff
@@ -49,4 +52,21 @@ describe('Spinner Specs', () => {
...
+	test('should apply custom aria-label to spinner', () => {
+		const customAriaLabel = 'custom aria label';
+		render(<Spinner aria-label={customAriaLabel} />);
+		const spinner = screen.getByRole('alert');
+
+		expect(spinner).toHaveAttribute('aria-label', customAriaLabel);
+	});
+
+	test('should render a spinner with \'blockClickOn\' = \'screen\' inside FloatingLayer', () => {
+		render(<FloatingLayerController><Spinner blockClickOn="screen" /></FloatingLayerController>);
```
jeonghee27, line 65: 👍🏻

Slider/tests/Slider-specs.js
jeonghee27: OK I understand what you intend.
Let's keep current

jeonghee27:
Good job! 
And I confirmed that the coverage cannot be increased anymore. 
Thank you for adding missing test cases even if they are not related to code cov increase.
Please check the a questions about "setTimeout", I'm not sure so please let me know which is correct

Slider/tests/Slider-specs.js
daniel-stoian-lgp: the NaN gave me headaches, to be honest. even with a single click on slider, the aria-valueText becomes NaN. I suspect some calculations of the new value cannot be computed in JSDom?
image
image

jeonghee27: We should defer seeing the mystery. :) If you want, you can create a separate issue and see deeper more or throw it to SWP.

daniel-stoian-lgp:
@jeonghee27 , I applied your suggestion. it works. However, i don't understand why Travis CI is still pending

PR #1325, opened 2022-09-21
WRO-12412: Fixed ImageItem to have proper size with importing VirtualList

MikyungKim:
### Checklist

ImageItem would have the wrong size of an image when it's horizontal with importing VirtualList or Dropdown like below.
image

### Resolution
The root cause of this issue is that VirtualList imports ImageItem.module.less in useThemeVirtualList.js to apply the scale class to ImageItem when `snapToCenter` is true.
When webpack bundles css, the sandstone/ImageItem/ImageItem.module.less is imported first for the above reason and then ui/Imageitem/ImageItem.module.less imported so the issue occurred.
To resolve this, I made `useThemeVirtualList.module.less` to have the scale class instead of importing ImageItem.module.less.

### Additional Considerations
With this change, the scale effect is going to apply to any ImageItem instead of "vertical, full image" only.

### Links
WRO-12412

### Comments

### DIFF
--- a/VirtualList/useThemeVirtualList.js
+++ b/VirtualList/useThemeVirtualList.js
@@ -13,7 +13,7 @@ import {useEventKey, useEventFocus} from './useEvent';
 import usePreventScroll from './usePreventScroll';
 import {useSpotlightConfig, useSpotlightRestore} from './useSpotlight';
 
-import ImageItemCss from '../ImageItem/ImageItem.module.less';
+import css from './useThemeVirtualList.module.less';
 
 const SpotlightAccelerator = new Accelerator();
 const SpotlightPlaceholder = Spottable('div');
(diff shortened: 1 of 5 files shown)


VirtualList/useThemeVirtualList.module.less
```diff
@@ -0,0 +1,14 @@
...
+
+.scaled {
+	transform: @sand-imageitem-vertical-focus-image-transform;
+}
+
+.applySkins({
+	.scaled {
+		box-shadow: @sand-imageitem-focus-bg-shadow;
+	}
+});
```
jeonghee27, line 14: Why don't we remove this styles on ImageItem.module.less?

VirtualList/useThemeVirtualList.js
```diff
@@ -13,7 +13,7 @@ import {useEventKey, useEventFocus} from './useEvent';
 import usePreventScroll from './usePreventScroll';
```
jeonghee27, line 13: I think we need change log

MikyungKim: I agree. I'll put it right before merge.

VirtualList/useThemeVirtualList.module.less
MikyungKim: 👍

jeonghee27 (approved):
LGTM

PR #1150, opened 2022-01-28
WRN-15665: VirtualList: Fix `HoverToScroll` to update properly after changing `dataSize`

jeonghee27:
Enact-DCO-1.0-Signed-off-by: Jeonghee Ahn (jeonghee27.ahn@lge.com)

### Checklist

* [x] I have read and understand the [contribution guide](http://enactjs.com/docs/developer-guide/contributing/)
* [x] A [CHANGELOG entry](http://enactjs.com/docs/developer-guide/contributing/changelogs/) is included
* [x] At least one test case is included for this feature or bug fix
* [x] Documentation was added or is not needed
* [x] This is an API breaking change

### Issue Resolved / Feature Added
[//]: # (Describe the issue resolved or feature added by this pull request)
VirtualList/VirtualGridList: hoverToScroll not working after rendering

### Resolution
[//]: # (Does the code work as intended?)
[//]: # (What is the impact of this change and *why* was it made?)

Cause : 
`HoverToScrollBase` hook function and life cycle functions should be invoked properly when `scrollContainer` is updated.
However, the lifecycle functions are called later by setState from update().
It is becuase `HoverToScrollBase` receive `scrollContainerHandle` instead of `scrollContainer`.

Resolution : 
I changed `HoverToScrollBase` to get `scrollContainerHandle.current` instead of `scrollContainerHandle`.

### Additional Considerations
[//]: # (How should the change be tested?)
[//]: # (Are there any outstanding questions?)
[//]: # (Were any side-effects caused by the change?)

### Links
[//]: # (Related issues, references)
WRN-15665

### Comments
Enact-DCO-1.0-Signed-off-by: Jeonghee Ahn (jeonghee27.ahn@lge.com)

PR #1336, opened 2022-10-05
WRO-13221: Fixed VirtualList to scroll properly by hover when scrollbar is hidden (2.0.x)

juwonjeong:
### Checklist

When horizontalScrollbar or verticalScrollbar is set to hidden scroll did not work with hover action in VirtualList.

### Resolution
To resolve the issue, we have to revert https://github.com/enactjs/sandstone/pull/1150.

In PR 1150, It fixed HoverToScrollBase to receive scrollContainerHandle.current instead of scrollContainerHandle.
However, this fix caused an error when the scrollbar is hidden, as VirtualList rendered only once in this case.

So we reverted PR 1150. And then we have to fix HoverToScroll to scroll properly by hover after dataSize is changed(after rendering). Add the props to the useLayoutEffect (line 177) dependency list so that the update() function can be called each time after rendering.

### Additional Considerations
Partially cherry-picked #1334 (8852fa9)

### Links
WRO-13221

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The following is a curated list of changes in the Enact sandstone module, newest
 ### Fixed
 
 - `sandstone/MediaPlayer.MediaControls` to focus properly when pressing up key from buttons after holding left or right keys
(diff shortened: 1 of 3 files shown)


MikyungKim (approved):
LGTM

PR #1192, opened 2022-03-11
WRN-13958: Update sandstone samples with new React style APIs

sjro:
Enact-DCO-1.0-Signed-off-by: SJ RO ([sj.ro@lge.com](mailto:sj.ro@lge.com))
### Checklist

* [x] I have read and understand the [contribution guide](http://enactjs.com/docs/developer-guide/contributing/)
* [ ] A [CHANGELOG entry](http://enactjs.com/docs/developer-guide/contributing/changelogs/) is included
* [x] At least one test case is included for this feature or bug fix
* [x] Documentation was added or is not needed
* [ ] This is an API breaking change

### Issue Resolved / Feature Added
[//]: # (Describe the issue resolved or feature added by this pull request)
The React style of samples is out of date.

### Resolution
[//]: # (Does the code work as intended?)
[//]: # (What is the impact of this change and *why* was it made?)
Update samples with new React style.

### Additional Considerations
[//]: # (How should the change be tested?)
[//]: # (Are there any outstanding questions?)
[//]: # (Were any side-effects caused by the change?)

### Links
[//]: # (Related issues, references)
WRN-13958

PR #1337, opened 2022-10-05
WRO-13165: Don't change the a11y setting when open qa-a11y readAlert menu

jeonghee27:
### Checklist

'READ ALERT' is read out when clicked on OK button.

### Resolution
The readAlert page itself is switching off and on accessibility. It is regression from https://github.com/enactjs/sandstone/pull/1192/files#diff-e371e57e16809e27da46201a26f7f79140ce83fc985c92183daec5d084a4decd

Since the setting value should not need to be changed when the page is loaded, I remove the wrong useEffect code.

### Additional Considerations

### Links
WRO-13165

### Comments

### DIFF
--- a/samples/qa-a11y/src/views/ReadAlert.js
+++ b/samples/qa-a11y/src/views/ReadAlert.js
@@ -34,24 +34,19 @@ const ReadAlertView = () => {
 		}
 	}, []);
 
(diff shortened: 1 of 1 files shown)


juwonjeong (approved):
LGTM

PR #1347, opened 2022-10-19
WRO-13139: Sampler: Remove controls from Docs tab

stanca-pop-lgp:

### Checklist

In storybook, if controls are changed from Docs tab, it doesn't update the component preview. 

### Resolution
Setting "inlineStories" to true causes bugs, so we agreed to keep inlineStories false and set a height to the iFrame. Also, controls were removed from the docs page.

### Additional Considerations

### Links
WRO-13139


adrian-cocoara-lgp (approved):
LGTM but let's see SWP team opinion

adrian-cocoara-lgp:
Travis fail because of node v19

samples/sampler/stories/default/About.js
```diff
@@ -64,7 +65,17 @@ const HintDialog = kind({
 });
 
 export default {
-	title: 'About/A Tour of Sampler'
+	title: 'About/A Tour of Sampler',
+	parameters: {
+		docs: {
```
MikyungKim, line 70: Oh... you don't need to do this for every story. You could just do it from the preview.js.

stanca-pop-lgp: Oh, I don't know why I didn't think of that. I will fix this. Thank you, Mikyung!

stanca-pop-lgp: I kept the imports sorting if that is ok.

samples/sampler/.qa-storybook/preview.js
```diff
@@ -48,12 +49,19 @@ const skins = {
...
 
 configureActions();
+
 export const parameters = {
 	docs: {
 		container: DocsContainer,
-		page: DocsPage,
-		iframeHeight: 360,
 		inlineStories: false,
+		iframeHeight: ri.scaleToRem(300),
```
MikyungKim, line 57: How about 600? It seems too small.

samples/sampler/.storybook/preview.js
```diff
@@ -47,12 +48,18 @@ const skins = {
...
 
 configureActions();
+
 export const parameters = {
 	docs: {
 		container: DocsContainer,
-		page: DocsPage,
-		iframeHeight: 360,
 		inlineStories: false,
+		iframeHeight: ri.scaleToRem(300),
```
MikyungKim, line 56: How about 900? It seems too small.

PR #1353, opened 2022-10-31
WRO-14597: Changed component to use VirtualList.

alexandrumorariu:
### Checklist

Changed the component to use VirtualList so it is more efficient.

### Resolution

### Additional Considerations

### Links
WRO-14597

### DIFF
--- a/TransferList/TransferList.js
+++ b/TransferList/TransferList.js
@@ -16,12 +16,12 @@ import {Cell, Layout} from '@enact/ui/Layout';
 import ri from '@enact/ui/resolution';
 import PropTypes from 'prop-types';
 import compose from 'ramda/src/compose';
-import {useCallback, useEffect, useRef, useState} from 'react';
+import {useCallback, useEffect, useMemo, useRef, useState} from 'react';
 
 import Button from '../Button';
 import CheckboxItem from '../CheckboxItem';
-import Scroller from '../Scroller';
 import Skinnable from '../Skinnable';
+import VirtualList from '../VirtualList';
 
 import componentCss from './TransferList.module.less';
 
@@ -79,6 +79,15 @@ const TransferListBase = kind({
 		height: PropTypes.number,
 
 		/**
+		 * The height of the checkbox item.
+		 *
+		 * @type {Number}
+		 * @default
+		 * @public
+		 */
+		itemSize: PropTypes.number,
+
+		/**
 		 * Allows items to be transferred from one list to another using Spotlight Right and/or Spotlight Left.
 		 *
 		 * @type {Boolean}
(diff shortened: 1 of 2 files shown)


daniel-stoian-lgp (approved):
LGTM

TransferList/TransferList.js
```diff
@@ -78,6 +78,15 @@ const TransferListBase = kind({
 		 */
 		height: PropTypes.number,
 
+		/**
+		 * The height of the checkbox item.
+		 *
+		 * @type {Number}
+		 * @default
```
adrian-cocoara-lgp, line 85:
```suggestion
		 * @default 201
```

TransferList/TransferList.js
```diff
@@ -128,59 +138,60 @@ const TransferListBase = kind({
...
+				<CheckboxItem
+					{...rest}
+					data-index={dataIndex}
+					draggable
+					className={componentCss.draggableItem}
+					id={`${index}-${list}`}
+					key={index + list}
+					onClick={handleClick}	// eslint-disable-line  react/jsx-no-bind
+					onSpotlightDown={handleSpotlightDown}	// eslint-disable-line  react/jsx-no-bind
+					onSpotlightUp={handleSpotlightUp}	// eslint-disable-line  react/jsx-no-bind
```
adrian-cocoara-lgp, line 172: add `react/jsx-no-bind` to the list on line 1 and remove them from here

samples/sampler/stories/default/TransferList.js
```diff
@@ -16,12 +16,14 @@ export const _TransferList = (args) => (
...
 		allowMultipleDrag={args['allowMultipleDrag']}
 		firstList={['Item1', 'Item2', 'Item3', 'Item4', 'Item5', 'Item6', 'Item7', 'Item8']}
+		itemSize={args['itemSize']}
 		moveOnSpotlight={args['moveElementOnSpotlightDirections']}
 		secondList={['Item9', 'Item10', 'Item11', 'Item12', 'Item13', 'Item14', 'Item15', 'Item16']}
 	/>
 );
 
 boolean('allowMultipleDrag', _TransferList, Config, true);
+number('itemSize', _TransferList, Config, 201);
```
adrian-cocoara-lgp, line 26: do we need to specify 201 here if we defined it as defaultProp?

alexandrumorariu: Just as a reference for the user.

TransferList/TransferList.js
alexandrumorariu: I would like to keep them like this. The default way to avoid this warning is to wrap the functions in useCallback, but as some of them do change on every rerender there is no point for it here. If we place it as a global rule it will not be enforced where we need it, so it would be better to do it locally.

adrian-cocoara-lgp: okay

adrian-cocoara-lgp (approved):
LGTM

PR #1361, opened 2022-11-22
WRP-3423: Add aria-own to prevent read parent page

jeonghee27:
### Checklist

 Tv reads out previous page's header after return the `PopupTabLayout` page from `Input` 's `Popup`

### Resolution
It is regression from https://github.com/enactjs/sandstone/pull/1289.
After removing "aria-hidden", the bug occurs.

So I add a wrapper `div` and set aria-own to prevent read parent page after return the previous page. 
It is exactly same way with `Alert` https://github.com/enactjs/sandstone/pull/813/

### Additional Considerations
Screenshot test pass : Jenkins enact-screenshot-tests/1086/

### Links
WRP-3423

### Comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The following is a curated list of changes in the Enact sandstone module, newest
 
 ### Fixed
 
+- `sandstone/Input` to read out properly after closing it in a `sandstone/PopupTabLayout`
 - `sandstone/ImageItem` to have proper size when imported with `sandstone/Dropdown` or `sandstone/VirtualList` in the same file
 - `sandstone/MediaPlayer.MediaControls` to focus properly when pressing up key from buttons after holding left or right keys
 - `sandstone/Scroller` and `sandstone/VirtualList` to scroll properly by hover when scrollbar is hidden or `dataSize` is changed
--- a/Input/Input.js
+++ b/Input/Input.js
@@ -394,65 +394,68 @@ const InputPopupBase = kind({
 		delete rest.onOpenPopup;
 
 		return (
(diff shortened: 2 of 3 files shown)


seunghoh (changes requested):
Would you please find out if there any better way? As you can see in the alert PR comment, and also in the https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-owns, `aria-owns` should NOT be used for parent/child DOM. 
This is a tricky workaround solution and I don't want to expand this workaround across more components.

jeonghee27:
> Would you please find out if there any better way? As you can see in the alert PR comment, and also in the https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-owns, aria-owns should NOT be used for parent/child DOM.
This is a tricky workaround solution and I don't want to expand this workaround across more components.


In JSX, it looks like a parent/children DOM, but in actual Rendered DOM trees, the two DOM are separate.
Because `Popup` redner `FloatingLayer` and `FloatLayer` attach a dom at `div with id="floatinglayer"` using `ReactDOM.createPortal`. Therefore, since the two DOM are separate that do not have a parent/children relationship at all, they need to be connected through aria-owns. So I don't think this method is tricky workaround, and it's an appropriate usage of aria-owns. That's why I it was used in Alert #813.

If you want avoid this method, we should revert #1289 to keep `aria-hidden`. But when using `aria-hidden`, I'm not sure why the headings aren't read when returning to the `popupTabLayout`.

seunghoh (approved):
LGTM

PR #1365, opened 2022-12-14
WRP-4977: Fixed MediaControls to disable hidden button properly

juwonjeong:
### Checklist

Buttons in the MediaControls were focusable and clickable even though it was hidden. 
They need to be disabled when they are hidden.

### Resolution
Added `visibility:hidden` css property to the related elements(moreButtonComponents, actionGuide).

### Additional Considerations

### Links
WRP-4977


0x64 (approved):
:lgtm:

PR #1374, opened 2022-12-28
WRP-5920: Fixed hoverToScroll not to invoke stopRaf when scrollContainer is updated

juwonjeong:

### Checklist

The issue was reported that when the pointer is in the hover area it scrolls for a while and then stopped in some cases.

### Resolution
We found that scrolls stopped because of stopRaf function in hoverToScroll.js invoked unintentionally. 
stopRaf function is invoked when scrollContainer is updated.

### Additional Considerations
We removed the code that the stopRaf function is called because of scrollerContainer update.
We found in hoverToScroll.js line 159 covers the edge case which needs to call the stopRaf function. 
So we decided to remove codes that call stopRaf function in a wide range of cases.
Later, we need to investigate whether scrollContainer is updated in unnecessary cases.

### Links
WRP-5920

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ The following is a curated list of changes in the Enact sandstone module, newest
 ### Fixed
 
 - `sandstone/MediaPlayer.MediaControls` to disable buttons when hidden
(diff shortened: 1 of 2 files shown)


0x64 (approved):
:lgtm:

PR #1389, opened 2023-01-26
WRP-6648:DayPicker: fix TypeError when passing prop 'selected' as a number type

hoeunryuLG:
### Checklist

DayPicker accepts a number or an array of numbers in "selected" prop.
But TypeError exception occurs in localizedSelected() function when passing a number (not 0).

localizedSelected() function also accepts a number or an array of numbers as a "selected" argument.
The problem is this.
 1. the function doesn't differentiate between number 0 and undefined/null.
 2. treats unconditionally the "selected" argument as an array.

### Resolution
The fix is this.
 1. differentiate between 0 and undefined/null
 2. check type of "selected" argument and if it's a number, then returns a localized number.
 3. leave the selected.map() for arrays as a fallback.

### Additional Considerations
Unit/Screenshot test has been added.

### Links
WRP-6648
QWTC-637

### Comments

### DIFF
--- a/DayPicker/DaySelectorDecorator.js
+++ b/DayPicker/DaySelectorDecorator.js
@@ -35,10 +35,15 @@ function generalizeSelected (selected, state) {
 	return selected.map(v => generalizeDay(v, state.firstDayOfWeek)).sort();
 }
 
-// Accepts a localized selected array and returns a "Sunday at index 0" array
+// Accepts a "Sunday at index 0" selected array or number and returns a
+// localized array or number.
 function localizeSelected (selected, state) {
-	if (state.firstDayOfWeek === 0 || !selected) {
+	if (selected === undefined || selected === null) {
 		return selected;
+	}
+
+	if (typeof selected === 'number') {
+		return localizeDay(selected, state.firstDayOfWeek);
 	}
 
 	return selected.map(v => localizeDay(v, state.firstDayOfWeek));
--- a/tests/screenshot/apps/components/DayPicker.js
+++ b/tests/screenshot/apps/components/DayPicker.js
@@ -26,6 +26,25 @@ const DayPickerTests = [
 		component: <DayPicker disabled selected={1} />
 	},
 
+	// *************************************************************
+	// locale = 'es-ES', [QWTC-637]
+	{
+		locale: 'es-ES',
+		component: <DayPicker />
+	},
+	{
+		locale: 'es-ES',
+		component: <DayPicker selected={1} />
+	},
+	{
+		locale: 'es-ES',
+		component: <DayPicker disabled />
+	},
+	{
+		locale: 'es-ES',
+		component: <DayPicker disabled selected={1} />
+	},
+
 	...withConfig({focus: true}, [
 		<DayPicker selected={2} />,
 		<DayPicker disabled selected={2} />
--- a/DayPicker/tests/DayPicker-specs.js
+++ b/DayPicker/tests/DayPicker-specs.js
@@ -25,6 +25,15 @@ describe('DayPicker', () => {
 		expect(selectedDay).toHaveClass(expected);
 	});
 
+	test('should select day when passed prop \'selected\' as a number', () => {
+		render(<DayPicker selected={0} />);
+		const selectedDay = screen.getAllByRole('checkbox')[1];
+
+		const expected = 'selected';
+
+		expect(selectedDay).toHaveClass(expected);
+	});
+
 	test('should emit an onSelect event with \'onSelect\' type when selecting days', () => {
 		const handleSelect = jest.fn();
 		render(<DayPicker onSelect={handleSelect} />);


DayPicker/DaySelectorDecorator.js
```diff
@@ -35,12 +35,17 @@ function generalizeSelected (selected, state) {
 	return selected.map(v => generalizeDay(v, state.firstDayOfWeek)).sort();
 }
 
-// Accepts a localized selected array and returns a "Sunday at index 0" array
+// Accepts a "Sunday at index 0" selected array or number and returns a
```
MikyungKim, line 38: Would you please explain why did you change this comment?
According to the first commit(https://github.com/enactjs/sandstone/pull/343/commits/a442055235a26bee21626b762776473ee5e2c04f), the input and output should not be changed.

DayPicker/DaySelectorDecorator.js
```diff
@@ -35,12 +35,17 @@ function generalizeSelected (selected, state) {
...
+// localized array or number.
 function localizeSelected (selected, state) {
-	if (state.firstDayOfWeek === 0 || !selected) {
+	if (typeof selected === 'undefined' || selected === null) {
```
MikyungKim, line 41: The original code had the intent to skip the "localizeDay" when the `firstDayOfWeek` is 0.
I don't see any reason for removing that condition.
How about
```suggestion
	if (state.firstDayOfWeek === 0 || selected == null) {
```

MikyungKim (commented):
Would you please add a changelog for this?

samples/sampler/stories/default/DayPicker.js
```diff
@@ -15,13 +15,15 @@ export const _DayPicker = (args) => (
 		<DayPicker
 			aria-label={args['aria-label']}
 			disabled={args['disabled']}
+			selected={args['selected']}
 			onSelect={action('onSelect')}
```
MikyungKim, line 19: Normally, we place props in alphabetical order.
```suggestion
			onSelect={action('onSelect')}
			selected={args['selected']}
```

samples/sampler/stories/default/DayPicker.js
```diff
@@ -15,13 +15,15 @@ export const _DayPicker = (args) => (
...
 		/>
 	</Scroller>
 );
 
 text('aria-label', _DayPicker, DayPicker);
 boolean('disabled', _DayPicker, DayPicker);
+number('selected', _DayPicker, DayPicker);
```
MikyungKim, line 26: `selected` can be either a number or an array. So this type seems wrong.
And normally this prop isn't set by a user. So I prefer not to have this prop in the default sampler but it seems okay for qa-sampler.

samples/sampler/stories/qa/DayPicker.js
```diff
@@ -3,6 +3,7 @@ import Heading from '@enact/sandstone/Heading';
 import Item from '@enact/sandstone/Item';
 import Scroller from '@enact/sandstone/Scroller';
 import {Component} from 'react';
+import {number} from '@enact/storybook-utils/addons/controls';
```
MikyungKim, line 6: As I was saying, would you please consider other types of controls for `selected`?
Maybe.. select?

MikyungKim: If you want to test "number" type only, it seems okay.

tests/screenshot/apps/components/DayPicker.js
```diff
@@ -26,6 +26,25 @@ const DayPickerTests = [
 		component: <DayPicker disabled selected={1} />
 	},
 
+	// *************************************************************
+	// locale = 'es-ES', [QWTC-637]
```
MikyungKim, line 30: @dongsuwon, Is the TC ticket number correctly related to these screenshot tests?

samples/sampler/stories/qa/DayPicker.js
```diff
@@ -48,3 +49,19 @@ ToTestGetSelectedDayString.parameters = {
...
+
+export const WithSelectedNumber = (args) => (
+	<Scroller>
+		<Heading size="small">
+			Change *selected* number.
+			<br />
+			Change locale to *es-ES* starting on Monday.
+		</Heading>
+		<Item label={getSelectedDayString(args['selected'])}>{'Selected Day'}</Item>
+		<DayPicker disabled selected={args['selected']} />
```
MikyungKim, line 61: Why is this component disabled?

MikyungKim (changes requested):
Thank you. I left some comments.

tests/screenshot/apps/components/DayPicker.js
dongsuwon: @MikyungKim 
Yes, It is to check parts that cannot be checked in ui-test.
e.g) color of check/uncheck state, color of disable state

hoeunryuLG (commented):
I'll update CHANGELOG.

DayPicker/DaySelectorDecorator.js
MikyungKim: Also, `selected == null` will return `true` when `selected` is `undefined` or `null`. You can use this.

samples/sampler/stories/qa/DayPicker.js
MikyungKim: I think it won't change the value when `selected` prop is given. Please check.

samples/sampler/stories/qa/DayPicker.js
MikyungKim: 🤔 I'm not sure. It seems okay to test only for the number type if we titled it well.
But if the case can be tested via Screenshot, do we need this sampler for sure?

DayPicker/DaySelectorDecorator.js
MikyungKim: 🤔 I think you are right. WDYT about the comment at line 28? Should we change that too?

hoeunryuLG: The comment on that line is also wrong. It describes precisely the opposite of the function's behavior.
The generalizeSelected function receives values that have already reflected the locale and convert them into values where Sunday is 0.
Comments could be changed to:
// Accept the localized array with "firstDayOfWeek" at index 0 and returns a "Sunday at index 0" array

DayPicker/DaySelectorDecorator.js
hoeunryuLG: I will use it to revise the conditional statement.

samples/sampler/stories/qa/DayPicker.js
hoeunryuLG: If you are unsure, I will remove the entire code change of qa-sampler.

DayPicker/DaySelectorDecorator.js
MikyungKim: 🆗

samples/sampler/stories/qa/DayPicker.js
MikyungKim: 🆗

samples/sampler/stories/qa/DayPicker.js
hoeunryuLG: I will remove the entire code change of qa-sampler

samples/sampler/stories/default/DayPicker.js
hoeunryuLG: I will remove the entire code change of sampler

hoeunryuLG:
1. Merge the latest develop branch
2. One fixup patch with the following changes
2.1 Add CHANGELOG entry
2.2 Return immediately if firstDayOfWeek === 0 in the localizeSelected function
2.3 Modify comment of generalizeSelected function
3. Modify the unit test so that it must go through the conditional statement that checks the number type.

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 The following is a curated list of changes in the Enact sandstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+- `sandstone/DaySelectorDecorator` to handle `number` type properly for prop `selected`,
```
MikyungKim, line 8: Since `DaySelectorDecorator` is a private component, we don't use this in the change log. Instead, we could use `sandstone/DayPicker`.

```suggestion
- `sandstone/DayPicker` to handle number typed `selected` prop properly in es-ES locale
```

hoeunryuLG: I've fixed it and pushed a fix-up patch.

MikyungKim (approved):
LGTM

seunghoh (approved):
💯

PR #1401, opened 2023-02-10
WRP-866: Migrate `TooltipDecorator` to hook-based implementation

0x64:
### Checklist

`TooltipDecorator` was not a hook-based one. Let's migrate it.

### Resolution
Created `useTooltip` and made `TooltipDecorator` to use it internally.

### Additional Considerations
I found that the default value of `TooltipDecorator`'s `screenEdgeKeepout` config was documented incorrectly. Looks simply missing when the default value was updated (#6). So, I fixed it also in this PR.

### Links
WRP-866


TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,266 @@
...
+
+	const startTooltipLayoutJob = useCallback(() => {
+		mutableRef.current.setTooltipLayoutJob?.startAfter(tooltipUpdateDelay);
+	}, [tooltipUpdateDelay]);
+
+	const handlers = useHandlers({
+		// Recalculate tooltip layout on keydown to make sure tooltip is positioned correctly in case something changes as a result of the keydown.
+		onKeyDown: handle(
+			forward('onKeyDown'),
+			() => (!disabled),
```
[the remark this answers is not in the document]
0x64: Due to the manner of setting the default value of _disabled_ prop in this hook, a suggested sentence could have a wrong behavior if a given _props_ has a _disabled_ prop with a value of _undefined_. But we already have a same pattern in _TabGroup.js_, so I will update this as the below.
> not(forProp('disabled', true))

TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,266 @@
...
+
+	useLayoutEffect(() => {
+		mutableRef.current.setTooltipLayoutJob = new Job(() => {
+			setTooltipLayout();
+		});
+	}, [setTooltipLayout]);
+
+	const getTooltipRef = useCallback((node) => {
+		if (node) {
+			tooltipRef.current = node;
```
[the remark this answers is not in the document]
0x64: Good catch! No, it's not fine. :) I will fix this.

TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,266 @@
...
+			}
+		};
+	}, [startTooltipLayoutJob]);
+
+	const setTooltipLayout = useCallback(() => {
+		if (!tooltipText || !tooltipRef.current || !clientRef.current) {
+			return;
+		}
+
+		const newState = getDirectionAnchor(tooltipPosition, tooltipType);
```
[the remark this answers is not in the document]
0x64: What about `newLayoutInfo`? This name can be paired with `setLayoutInfo`.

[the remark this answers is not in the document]
juwonjeong: Okay, It looks good to me.

juwonjeong (approved):
LGTM

TooltipDecorator/TooltipDecorator.js
```diff
@@ -39,10 +30,10 @@ const defaultConfig = {
 	 * resolution.
 	 *
 	 * @type {Number}
-	 * @default 24
+	 * @default 48
 	 * @memberof sandstone/TooltipDecorator.TooltipDecorator.defaultConfig
 	 */
-	screenEdgeKeepout: (24 + 24),
+	screenEdgeKeepout: (24 + 24), // Do NOT forget to check useTooltip's default prop value also.
```
MikyungKim, line 36: 🤔 Do we really need this comment....?

TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,266 @@
...
+import {FloatingLayerBase} from '@enact/ui/FloatingLayer';
+import ri from '@enact/ui/resolution';
+
+import {Tooltip, defaultArrowAnchor, defaultDirection} from './Tooltip';
+import {adjustDirection, adjustAnchor, calcOverflow, getLabelOffset, getPosition} from './util';
+
+let currentTooltip; // needed to know whether or not we should stop a showing job when unmounting
+
+import useHandlers from '@enact/core/useHandlers';
+import {useCallback, useEffect, useLayoutEffect, useRef, useState} from 'react';
```
MikyungKim, line 15: Why is this imports are here?

TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,266 @@
...
+
+function getDirectionAnchor (tooltipPosition, tooltipType) {
```
MikyungKim, line 17:
```suggestion
function getTooltipDirection (tooltipPosition, tooltipType) {
```

MikyungKim: Let's convert this to an arrow function.

TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,266 @@
...
+		};
+	} else {
+		return {
+			tooltipDirection: position === 'below' ? 'below' : 'above',
+			arrowAnchor: 'right'
+		};
+	}
+}
+
+const removeTooltipProps = ({...props}) => {
```
MikyungKim, line 34: Why is the parameter `{...props}`? Can we just use `props`?

TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,266 @@
...
+	delete props.tooltipText;
+	delete props.tooltipType;
+	delete props.tooltipUpdateDelay;
+	delete props.tooltipWidth;
+
+	return props;
+};
+
+// A hook to show Sandstone-styled tooltip components.
+function useTooltip (props = {}) {
```
MikyungKim, line 51: Why do we need the default value for `props`?

MikyungKim: I slightly prefer to be an arrow function.

TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,266 @@
...
+	const {
```
MikyungKim, line 52: Would you please add propTypes for the props?

MikyungKim: Let's define these props one in a line.

TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,266 @@
...
+		screenEdgeKeepout = (24 + 24), // Do NOT forget to check TooltipDecorator's default config value also.
+		tooltipDelay = 500, tooltipType = 'balloon', tooltipUpdateDelay = 400,
+		tooltipMarquee, tooltipPosition, tooltipProps, tooltipRelative, tooltipText, tooltipWidth
+	} = props;
+	const rtl = useI18nContext()?.rtl;
```
MikyungKim, line 57: `rtl` can be provided via prop from `TooltipDecorator` right?
Would you please consider that case too?

TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,266 @@
...
+
+	const [showing, setShowing] = useState(false);
+	const [layoutInfo, setLayoutInfo] = useState({
```
MikyungKim, line 60: Let's remove `Info` from the name.

TooltipDecorator/TooltipDecorator.js
MikyungKim, line 36: Why don't we remove the comment here and make a constant from `useTooltip` to share the value here?

TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,266 @@
...
+		onBlur: handle(
+			forward('onBlur'),
+			hideTooltip
+		)
+	}, props);
+
+	useEffect(() => {
+		const mutableRefCurrent = mutableRef.current;
+
+		if (window.MutationObserver) {
```
MikyungKim, line 142: Please check window existence with `typeof`.

TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,266 @@
...
+				currentTooltip = null;
+
+				mutableRefCurrent.mutationObserver?.disconnect();
+				mutableRefCurrent.resizeObserver?.disconnect();
+
+				mutableRefCurrent.showTooltipJob.stop();
+				mutableRefCurrent.setTooltipLayoutJob.stop();
+			}
+		};
+	}, [startTooltipLayoutJob]);
```
MikyungKim, line 161: Shouldn't creating and disconnecting `mutationObserver` and `resizeObserver` be for only mounting and unmounting?

TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,266 @@
...
+
+	const setTooltipLayout = useCallback(() => {
```
MikyungKim, line 163: Where is the code relevant to https://github.com/enactjs/sandstone/blob/master/TooltipDecorator/TooltipDecorator.js#L270?

TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,266 @@
...
+		if (!tooltipText || !tooltipRef.current || !clientRef.current) {
```
MikyungKim, line 164: Would you explain why `!tooltipText` added here?

TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,266 @@
...
+			return;
+		}
+
+		const newLayoutInfo = getDirectionAnchor(tooltipPosition, tooltipType);
```
MikyungKim, line 168: Let's remove `Info` from here.

TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,266 @@
...
+
+		const tooltipNode = tooltipRef.current.getBoundingClientRect(); // label bound
+		const clientNode = clientRef.current.getBoundingClientRect(); // client bound
+		const overflow = calcOverflow(tooltipNode, clientNode, newLayoutInfo.tooltipDirection, ri.scale(screenEdgeKeepout));
+
+		newLayoutInfo.tooltipDirection = adjustDirection(newLayoutInfo.tooltipDirection, overflow, rtl);
+		newLayoutInfo.arrowAnchor = adjustAnchor(newLayoutInfo.arrowAnchor, newLayoutInfo.tooltipDirection, overflow, rtl);
+		newLayoutInfo.position = getPosition(clientNode, newLayoutInfo.tooltipDirection);
```
MikyungKim, line 176: Let's not rename this to `position`. Let's use `tooltipPosition`.

TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,266 @@
...
+			(newLayoutInfo.labelOffset !== layoutInfo.labelOffset) ||
+			(newLayoutInfo.arrowAnchor !== layoutInfo.arrowAnchor)
+		) {
+			setLayoutInfo(newLayoutInfo);
+		}
+	}, [tooltipText, rtl, tooltipPosition, tooltipType, screenEdgeKeepout, layoutInfo, tooltipRef, clientRef]);
+
+	useLayoutEffect(() => {
+		mutableRef.current.showTooltipJob = new Job(() => {
+			if (!showing) {
```
MikyungKim, line 191: Let's remove this if.

TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,266 @@
...
+			'--tooltip-position-left': tooltipRelative ? null : ri.unit(left, 'rem')
+		};
+
+		/**
+		 * Conditionally creates the FloatingLayer and Tooltip based on the presence of
+		 * `tooltipText` and returns a property bag to pass onto the Wrapped component
+		 *
+		 * @returns {Object} Prop object
+		 * @private
+		 */
```
MikyungKim, line 224: Should this comment be line 209?

TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,266 @@
...
+			);
+		} else if (showing) {
+			return renderedTooltip;
+		} else {
+			return null;
+		}
+	}, [getTooltipRef, hideTooltip, layoutInfo.arrowAnchor, layoutInfo.labelOffset, layoutInfo.position, layoutInfo.tooltipDirection, showing, tooltipMarquee, tooltipProps, tooltipRelative, tooltipText, tooltipType, tooltipWidth]);
+
+	return {
+		tooltipChildren: tooltipText ? renderTooltip() : null,
```
MikyungKim, line 257: Would you explain why you didn't add `tooltipDestinationProp` to useTooltip?

TooltipDecorator/useTooltip.js
0x64: I'm wondering, too. Why did I do like this?

TooltipDecorator/useTooltip.js
0x64: 👍

TooltipDecorator/useTooltip.js
0x64: This function's role is to delete useless properties from immutable `props`. If the name 'props' makes misunderstanding, I can rename it to 'rest' or something other. Any suggestion?

TooltipDecorator/useTooltip.js
0x64: If there is no default value for `props`, I need to check if it is a valid object before destructure it to pull out properties in it. So, simply for clear code.
Well, normal apps will pass an object normally at least for `tooltipText`, so we may remove the default value dealing with `props` as required argument.
I will change it to an arrow function.

TooltipDecorator/useTooltip.js
0x64: This is a hook, so propTypes not working. I will update props to be one in a line.

TooltipDecorator/useTooltip.js
0x64: Yes, `rtl` is passed as a part of TooltipDecorator.`rtl` is pased. And, no, we can't assume that as a standalone hook. That is why I use `useI18nContext` here.

TooltipDecorator/useTooltip.js
0x64: 🆗

TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,266 @@
...
+	const rtl = useI18nContext()?.rtl;
+
+	const [showing, setShowing] = useState(false);
+	const [layoutInfo, setLayoutInfo] = useState({
+		arrowAnchor: null,
+		labelOffset: null,
+		position: {top: 0, left: 0},
+		tooltipDirection: null
+	});
+	const mutableRef = useRef({
```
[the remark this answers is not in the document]
0x64: 👍

TooltipDecorator/useTooltip.js
0x64: `tooltipPosition` is the name of a prop and its value is string like 'above', 'left middle', etc. `position` is a calculated number value. The original name leads misunderstanding, so I intentionally changed it.

TooltipDecorator/useTooltip.js
0x64: Ah, right. Thanks!

TooltipDecorator/useTooltip.js
0x64: You're right. My mistake.

TooltipDecorator/useTooltip.js
0x64: As a HOC, TooltipDecorator must get the name of a prop passed to a wrapped component since HOC does not know a wrapped component's props.
But, when a parent functional component or a parent(?) hook uses `useTooltip`, a parent knows the name of the right prop and handles the prop directly. Just `tooltipChildren` is required.

TooltipDecorator/useTooltip.js
0x64: Well, not remember whole my thoughts during the work, I propabely add a dependency simply to turn off the warning (or pushed a button to update automatically by IDE).
Note that is is not a must-do thing to connect and disconnect only mounting and unmounting since the current code works safely.

TooltipDecorator/useTooltip.js
0x64: Good catch! I missed this. Thank you!

TooltipDecorator/useTooltip.js
0x64: 🆗

TooltipDecorator/TooltipDecorator.js
0x64: 🆗

TooltipDecorator/useTooltip.js
0x64: I removed the default value for `props` after discussion with Mikyung.

TooltipDecorator/useTooltip.js
0x64: I updated to use `props.rtl` if a boolean value is given after discussion with Mikyung as it is a natural usage.

TooltipDecorator/useTooltip.js
0x64: Concluded to keep the current dependency.

TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,281 @@
...
+		tooltipProps,
+		tooltipRelative,
+		tooltipText,
+		tooltipWidth
+	} = props;
+
+	let rtl = useI18nContext()?.rtl;
+	if (typeof props.rtl === 'boolean') {
+		rtl = props.rtl;
+	}
```
MikyungKim, line 68: How about 
```suggestion
	let rtl = props.rtl ? props.rtl : useI18nContext()?.rtl;
```

TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,281 @@
...
+	}, [tooltipUpdateDelay]);
+
+	// Recalculate tooltip layout on keydown to make sure tooltip is positioned correctly in case something changes as a result of the keydown.
+	const onKeyDown = useCallback((ev) => {
+		forward('onKeyDown', ev, props);
+		if (!props.disabled) {
+			startTooltipLayoutJob();
+		}
+	}, [props, startTooltipLayoutJob]);
+	const onMouseOver = useCallback((ev) => {
```
MikyungKim, line 123: Let's have an empty line between consts.

TooltipDecorator/useTooltip.js
0x64: Cannot call `useI18nContext` conditionally.

TooltipDecorator/useTooltip.js
0x64: 🆗

TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,283 @@
...
+	delete props.tooltipRelative;
+	delete props.tooltipText;
+	delete props.tooltipType;
+	delete props.tooltipUpdateDelay;
+	delete props.tooltipWidth;
+
+	return props;
+};
+
+const defaultScreenEdgeKeepout = (24 + 24);
```
MikyungKim, line 48: How about adding the same explanation as `screenEdgeKeepout` here?

TooltipDecorator/useTooltip.js
```diff
@@ -0,0 +1,290 @@
...
+		if (node) {
+			setTooltipLayout();
+		}
+	}, [setTooltipLayout]);
+
+	/**
+	 * Conditionally creates the FloatingLayer and Tooltip based on the presence of
+	 * `tooltipText` and returns a property bag to pass onto the Wrapped component
+	 *
+	 * @returns {Object} Prop object
```
MikyungKim, line 237: 🤔 I guess it should be changed to proper one.

MikyungKim (approved):
LGTM

PR #1405, opened 2023-02-14
WRP-8078:Button: Add roundBorder prop making both sides of button fully round

hoeunryuLG:
### Checklist

As a result of a discussion with the GUI team, it was decided to provide a boolean prop to have both the left and right sides of the button completely round. Therefore, a `roundBorder` boolean prop is added to achieve that.

### Resolution
Use `border-radius: 9999px`.
    
* https://w3c.github.io/csswg-drafts/css-backgrounds/#corner-overlap
    
Corner curves must not overlap: When the sum of any two adjacent border radii exceeds the size of the border box, UAs must proportionally reduce the used values of all border radii until none of them overlap.
    
You can also use `border-radius: calc($height / 2)` for the exact height, but that will require a .sand-custom-text mixin nested inside .applySkins mixin for the "large text" environment.

    .sand-custom-text({
        &.roundBorder {
            .bg {
                 border-radius: calc(@sand-button-height-large / 2);
            }
        }
    });
    
To avoid this, we use the property of CSS specification that corner curves do not overlap when the border-radius is set to a tremendous value.

### Additional Considerations

### Links
WRP-8078

### Comments

### DIFF
--- a/Button/Button.js
+++ b/Button/Button.js
@@ -167,6 +167,15 @@ const ButtonBase = kind({
 		minWidth: PropTypes.bool,
 
 		/**
+		 * True if both sides of button are fully rounded.
+		 *
+		 * @type {Boolean}
+		 * @default false
+		 * @public
+		 */
+		roundBorder: PropTypes.bool,
+
+		/**
 		 * The size of the button.
 		 *
 		 * @type {('large'|'small')}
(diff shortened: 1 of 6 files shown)


hoeunryuLG:
This will be replaced with a totally different implementation approach.

PR #1417, opened 2023-03-07
WRP-6257: Layout: Added screenshot tests for `grow` prop of `Cell`

hoeunryuLG:
### Checklist

A follow-up PR to add qa-sampler and screenshot tests for `grow` prop in `Cell` (enactjs/enact#3128)

### Resolution
- ~~Added qa-sampler for `Layout`~~
- Added screenshot tests

~~qa-sampler~~
~~- One Column and Nested three Rows with a total of 9 Cells to control them at once.~~
~~- Three Cells in a Layout to control them individually.~~

~~In principle, this should be added to enact, but since enact does not have qa-sampler, it is added to sandstone instead.~~

screenshot tests

* Row
  - Simply set the `grow` prop to check if each cell occupies the same area.
  - Set `grow` and `size={1000} to make sure it fills the remaining area of the container.
  - by setting `grow` and `size={1500}`
     - Check that it does not shrink below the specified size
     - When `wrap` is activated, cells are listed in two lines and check if the last cell fills up the container.
    
* Column
  - Simply set the `grow` prop to check if each cell occupies the same area.
  - Set `grow` and `size={100}` to make sure it fills the remaining area of the container.
  - Set `grow` and `size={500}` to make sure it doesn't shrink below the
      specified size.

* Layout
  - by setting `grow` and `size={1500}`
     - Check that it does not shrink below the specified size.
     - When `wrap` is activated, check if cells are listed in two lines, and the last cell fills up the container.
 
### Additional Considerations
enactjs/enact#3128 should be merged first.

### Links
WRP-6257
enactjs/enact#3128

### Comments


samples/sampler/stories/qa/Layout.js
```diff
@@ -0,0 +1,136 @@
+import {boolean, select} from '@enact/storybook-utils/addons/controls';
+import Layout, {Cell, Column, Row} from '@enact/ui/Layout';
+
+export default {
+	title: 'Sandstone/Layout',
```
MikyungKim, line 5: I'm not sure about adding this to sandstone qa-sampler. 🤔
How about adding `grow` prop to the existing enact ui/Layout sampler?

tests/screenshot/apps/components/Layout.js
```diff
@@ -41,7 +41,25 @@ const rowTestCases = [
...
+	</Layout>,
+
+	<Row>
+		<Cell className={css.red} grow>Layout Head</Cell>
+		<Cell className={css.blue} grow>Layout Content</Cell>
+		<Cell className={css.green} grow>Layout Tail</Cell>
+	</Row>,
+
+	<Row>
+		<Cell className={css.red} grow size={1000}>Layout Head</Cell>
```
MikyungKim, line 53: The above Row result and this Row result look the same. Is this intended? Or should we change the size..?

tests/screenshot/apps/components/Layout.js
```diff
@@ -76,6 +94,24 @@ const columnTestCases = [
...
+	</Column>,
+
+	<Column style={{height: "500px"}}>
+		<Cell className={css.red} grow>Column header6</Cell>
+		<Cell className={css.blue} grow>Column Content6</Cell>
+		<Cell className={css.green} grow>Column footer6</Cell>
+	</Column>,
+
+	<Column style={{height: "500px"}}>
+		<Cell className={css.red} grow size={100}>Column header7</Cell>
```
MikyungKim, line 106: Likewise, the previous one and this test look similar.

samples/sampler/stories/qa/Layout.js
hoeunryuLG: I thought sampler and qa-sampler had different purposes.
The sampler had a showcase for UI components, and the qa-sampler had various test cases for UI components. But Enact doesn't have a qa-sampler, so I put this code in Sandstone's qa-sampler. But I will if you want to place this code into Enact's sampler.

tests/screenshot/apps/components/Layout.js
hoeunryuLG: This is intended. The result of both is the same. However, I want to test that when the sum of the sizes of three Cells is smaller than the size of the container, the three Cells are arranged side by side and fill the remaining space of the container.

tests/screenshot/apps/components/Layout.js
hoeunryuLG: Same as the intent of the test above.

samples/sampler/stories/qa/Layout.js
MikyungKim: If we could cover the test with Screenshot tests, I think that is enough.
Let's not add this and see if you can add the `grow` prop to the default Layout sampler from enact sampler.

hoeunryuLG:
The PR has been updated to address reviewer's comments.

MikyungKim (approved):
LGTM

PR #1425, opened 2023-03-17
WRP-10826: Fix `WizardPanels` to restore focus properly after a transition

hoeunryuLG:
### Checklist

Until 2.5.8, the focus always exists on the screen in 5-way mode when the next/prev button of the WizardPanels sample is selected. From 2.6.0, focus disappears when moving from view 3 to view 4 via the next button and from view 2 to view 1 via the prev button.

This behavior is observed on PC, and the ui-test on the local machine may fail. The failed test case is described as "should select buttons over header - [QWTC-2390]" in WizardPanels specs. Note that we have yet to get this issue on TV and Jenkins tests.

This problem is observed in Chrome version 104 or higher but not in Chrome version 103 or lower.

The observed problems are as follows: Spotlight refocuses on the item previously focused on in the previous view.
- When transitioning from View 3 to View 4, the focus remains on Item (Hello Item) in View 3.
- When transitioning from View 2 to View 1, the focus remains on Button A in View 2.

The problem goes away if adding some delay in `setTimeout()` in the `onTransition` Handler. There is a race condition between view transition and rendering.

### Resolution
Add 40ms delay to setTimeout in onTranstion
The delay of "40ms" ensures that the timer handler is called after two rendering cycles (60HZ).

### Additional Considerations
It should be checked to see if the problem is resolved in the sampler and ui-test on local machine with Chrome version 104 or higher.

### Links
WRP-10826
QWTC-2390

### Comments
Enact-1.0-DCO-Signed-off-by: Hoeun Ryu <hoeun.ryu@lge.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The following is a curated list of changes in the Enact sandstone module, newest
 
 ### Fixed
 
-- `sandstone/WizardPanels` to restore focus properly after a transtion
+- `sandstone/WizardPanels` to restore focus properly after a transition
 
 ## [2.6.3] - 2023-03-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The following is a curated list of changes in the Enact sandstone module, newest
 
 ### Fixed
 
-- `sandstone/WizardPanels` to handle focus properly when switching views in Chrome version 104 or higher
+- `sandstone/WizardPanels` to restore focus properly after a transtion
 
 ## [2.6.3] - 2023-03-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact sandstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `sandstone/WizardPanels` to handle focus properly when switching views in Chrome version 104 or higher
+
 ## [2.6.3] - 2023-03-17
 
 ### Added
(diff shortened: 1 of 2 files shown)


juwonjeong:
Could you please add a changelog?

hoeunryuLG:
I'll add an entry to CHANGELOG.

hoeunryuLG:
The PR has been updated to address comments.
- add a Fixed entry to CHANGELOG
- merged with the latest develop branch

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact sandstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `sandstone/WizardPanels` to handle focus properly when switching views in Chrome version 104 or higher
```
juwonjeong, line 9:
```suggestion
- `sandstone/WizardPanels` to restore focus properly after a transition in Chrome 104 or higher
```
I would like to discuss how to update the changelog with merger @0x64.

hoeunryuLG:
CHANGELOG has been update.

hoeunryuLG:
fix typo

WizardPanels/useFocusOnTransition.js
```diff
@@ -12,7 +12,7 @@ const transitionHandlers = {
 				if (spotlightId && !currentSpotlight) {
 					Spotlight.focus(spotlightId);
 				}
-			});
+			}, 40);
```
seunghoh, line 15: Is this the only way we can do?

hoeunryuLG: We don't have a handler that is notified when all child components finish rendering.
I believe this is necessary to ensure that the rendering cycles are completed, allowing Spotlight to navigate to a target within the child components.

0x64 (approved):

PR #1442, opened 2023-04-25
WRP-13074: call getLastInputType method only when it's available

hoeunryuLG:
### Checklist

This issue was raised because we unconditionally execute `getLastInputType` LS call without checking for the existence of the function. So, `surface-manager` reports an `LS_NO_METH` error on RP-based platforms.

### Resolution
The `getLastInputType` method only exists on TV platforms, not RP or OSE. So we need to check the function's existence first before calling the function.

We can query the API method list using a special method.
- `com/palm/luna/private/introspection`

Using this API, ThemeDecorator checks the existence of `getLastInputType` function in `constructor` and calls the function only when it is available in `componentDidMount`.

Please take a look at the commit message for more details. [9d35bbb](https://github.com/enactjs/sandstone/commit/9d35bbb7c03a078b96c8491cfe556a5edef667db)
### Additional Considerations
`surface-manager` should not report an `LS_NO_METH` error anymore on RP-based platforms.

### Links
WRP-13074

### Comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The following is a curated list of changes in the Enact sandstone module, newest
 
 - `sandstone/Picker` to include `type` in the event payload for `onChange`
 - `sandstone/Scroller` and `sandstone/VirtualList` to handle focus properly via page up at the first page and page down at the last page
+- `sandstone/ThemeDecorator` to call the API checking the last input type only when it's available
 - `sandstone/WizardPanels` to restore focus properly after a transition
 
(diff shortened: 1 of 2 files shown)


hoeunryuLG:
codecov drops by 0.03, but LS2Request code is not a component we can write a unit test for.

hoeunryuLG:
The PR has been merged with the latest develop branch

hoeunryuLG:
The approach of asking whether a function exists and then calling it cannot be used. Using the mentioned private LS API is a problem, and Enact has no choice but to call the function once.

PR #1452, opened 2023-05-24
WRP-15428: Add hiding element feature to EditableScroller

juwonjeong:
### Checklist

Add hiding/showing element feature when scroller is in edit mode(when `editable` is given)

### Resolution
- Item will be selected by single tap/click
- The selected item can be hidden and also hidden items can be shown
- Added `longPresMode` feature to `editable` object, allowing long-press or single tap to select an item.
- Fixed a bug where removing a selected item while moving was not working properly.
- Added qa-sampler 

### Additional Considerations

### Links
WRP-15428

### DIFF
--- a/Scroller/EditableWrapper.js
+++ b/Scroller/EditableWrapper.js
@@ -394,6 +394,7 @@ const EditableWrapper = (props) => {
 		const {prevToIndex, selectedItem} = mutableRef.current;
 
 		if (selectedItem) {
+			// rearrangedItems need for the case when removing item while moving selected item
 			const rearrangedItems = mutableRef.current.rearrangedItems;
 			const selectedItemRect = selectedItem && selectedItem.getBoundingClientRect();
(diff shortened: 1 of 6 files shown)


Scroller/EditableWrapper.js
```diff
@@ -40,7 +40,10 @@ const EditableShape = PropTypes.shape({
 	onComplete: PropTypes.func.isRequired,
 	centered: PropTypes.bool,
 	css: PropTypes.object,
-	removeItemFuncRef: EnactPropTypes.ref
+	hideItemFuncRef: EnactPropTypes.ref,
+	longPressMode: PropTypes.bool,
```
0x64, line 44: Let's name it `selectItemBy` in string type to make it clear.

juwonjeong: It's fixed.

0x64 (approved):

PR #1464, opened 2023-06-23
WRP-19408: Added `QuickGuidePanels` prop `onClose`

0x64:
### Checklist

`onClose` event handler prop is missing for `QuickGuidePanels`.

### Resolution
Added `onClose` prop

### Additional Considerations
Note that aria label is not defined yet, so it should be updated later.

A unit test is added.

### Links
WRP-19408

### Comments

### DIFF
--- a/QuickGuidePanels/QuickGuidePanels.js
+++ b/QuickGuidePanels/QuickGuidePanels.js
@@ -28,6 +28,15 @@ const QuickGuidePanels = (props) => <WizardPanels fullScreenContent {...props} /
 
 QuickGuidePanels.propTypes = {
 	/**
+	 * Hint string read when focusing the close button.
+	 *
+	 * @type {String}
+	 * @default 'Exit quick guide'
+	 * @private
+	 */
+	closeButtonAriaLabel: PropTypes.string,
+
+	/**
 	 * Called when the index value is changed.
 	 *
 	 * @type {Function}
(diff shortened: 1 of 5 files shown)


juwonjeong (approved):
LGTM

QuickGuidePanels/QuickGuidePanels.js
```diff
@@ -27,6 +27,15 @@ import WizardPanels from '../WizardPanels';
 const QuickGuidePanels = (props) => <WizardPanels fullScreenContent {...props} />;
 
 QuickGuidePanels.propTypes = {
+	/**
+	 * Hint string read when focusing the close button.
+	 *
+	 * @type {String}
+	 * @default 'Exit quick guide'
```
seunghoh, line 34: Don't we need to wrap by $L?

0x64: https://github.com/enactjs/sandstone/pull/1464/files#diff-4313679dd9c05a6dd7acdfaf794a2b0f0184323b7c8b384a3fed24533f2ed13dR479

There is $L already. For docs, some components describe $L also, but some components don't.

seunghoh: Yes, that is my question. Shall we use &L on docs or not :)

0x64: 🤖 For now, the same manner as other "panels". In general, we need to talk.

PR #1470, opened 2023-07-11
WRP-20570: `hoverToScroll` works properly inside another scroller

0x64:
### Checklist

Scrolling by `hoverToScroll` stops if *ALL* conditions below meet,
- A scroller or list (inner scroller) exists inside another scroller (outer scroller) that has a orthogonal scroll direction to the inner scroller.
- The inner scroller has `hoverToScroll` prop.
- The outer scroller has a (always either during-scroll) visible scrollbar.
- A user moves the pointer on a cross-area of "hoverToScroll" area of the inner scroller and "scrollbar" area of the outer scroller.

This pull request is to make the outer scroller scrolling properly in the situation described above.

### Resolution

The cause is that "scrollbar" DOM of the outer scroller is layered above the inner scroller in the cross-area, So, pointer events are dispatched to the "scrollbar" DOM. These events cannot be propagated to the inner scroller ever due to its DOM tree structure.

To make events to be dispatched to the inner scroller nodes, "pointer-events: none" CSS attribute is added to the "scrollbar" DOM of the outer scroller.

### Additional Considerations

### Links
WRP-20570

### Comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact sandstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `sandstone/Scroller` and `sandstone/VirtualList` to not stop scrolling by `hoverToScroll` inside another scroller
+
 ## [2.7.2] - 2023-06-30
 
 ### Added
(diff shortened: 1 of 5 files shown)


juwonjeong (approved):
LGTM

PR #1483, opened 2023-08-07
WRP-22553: Fix Dropdown to focus properly the first option and the last option via page up and page down

0x64:
### Checklist

Focus disappears on the first options and the last options when pressing page up/down on Dropdown that has more than 5 options.

### Resolution
The issue occurs after #1437. The cause is that a logic uses DOM node rather than Spotlight container ID to find Spotlight target after scrolling from the same position as before scrolling. To avoid issue of #1437, I updated the logic to use Spotlight container ID of the same DOM node the logic used instead of revert the PR.

### Additional Considerations
UI tests are added.

### Links
WRP-22553
#1437

### Comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact sandstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `sandstone/Dropdown' to focus properly the first option and the last option via page up and page down
+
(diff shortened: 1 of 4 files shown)


0x64:
Note: codecov result is not reported due to unknown reason. Here is the report: https://app.codecov.io/gh/enactjs/sandstone/pull/1483

juwonjeong (approved):
LGTM

juwonjeong (approved):
LGTM

PR #1490, opened 2023-08-23
WRP-24606: Fix hint dialog position of About page in the Sampler

vJIYEv:
### Checklist

In 'About' page of sandstone sampler, hint dialog line looks broken when changing the sampler view height.

### Resolution
A hint dialog should have either `top` or `bottom` property.
But broken hint dialogs have both of them.
So, delete `top` property of broken hint dialogs.

### Additional Considerations

### Links
WRP-24606

### Comments


samples/sampler/stories/default/About.js
```diff
@@ -85,7 +85,7 @@ export const ATourOfSampler = (args) => (
 		>
 			Select any component from the <b>sidebar</b> to see how it works
 		</HintDialog>
-		<HintDialog style={{bottom: riSafe(edgeDotKeepout), left: 39, top: 480}} length={540}>
+		<HintDialog style={{bottom: riSafe(edgeDotKeepout), left: 39}} length={348}>
```
seunghoh, line 88: How does come the 348 value? 
Why only bottom wrapped by riSafe? Why left is not wrapped?

samples/sampler/stories/default/About.js
```diff
@@ -95,7 +95,7 @@ export const ATourOfSampler = (args) => (
 			</b>{' '}
 			for a demonstration
 		</HintDialog>
-		<HintDialog style={{bottom: riSafe(edgeDotKeepout), left: 114, top: 582}} length={348}>
+		<HintDialog style={{bottom: riSafe(edgeDotKeepout), left: 114}} length={156}>
```
seunghoh, line 98: How does come the 156 value? 
Why only bottom wrapped by riSafe? Why left is not wrapped?

samples/sampler/stories/default/About.js
vJIYEv: 348 is a heuristic value that makes hint dialogs do not overlap each other.
I refered prior length of line 98. The hint dialog of line 98 didn't overlap with the one of line 81.
[bottom] should be wapped by riSafe because the size of hint dialog changes when you expand the side bar left to right.
But [left] position should not because the position of [Actions] and [Controls] buttons maintains their size.

samples/sampler/stories/default/About.js
vJIYEv: 156 is a heuristic value that makes hint dialogs do not overlap each other.
I refered prior difference of the hint dialog length between line 88 and line 98.
[bottom] should be wapped by riSafe because the size of hint dialog changes when you expand the side bar left to right.
But [left] position should not because the position of [Actions] and [Controls] buttons maintains their size.

samples/sampler/stories/default/About.js
juwonjeong: As I tested, hint dialogs still overlap each other. Could you please check if it can be fixed?

juwonjeong (approved):

seunghoh (approved):
💯

PR #1499, opened 2023-09-06
WRP-26268: Rename PalmServiceBridge as WebOSServiceBridge

0x64:
### Checklist

`PalmServiceBridge` was renamed as `WebOSServiceBridge` in 2018 with an fallback alias for compatibility on legacy webOS.

### Resolution
Updated qa-a11y sample to use `WebOSServiceBridge`

### Additional Considerations

### Links
WRP-26268

### Comments

### DIFF
--- a/samples/qa-a11y/src/views/ReadAlert.js
+++ b/samples/qa-a11y/src/views/ReadAlert.js
@@ -18,7 +18,7 @@ const ReadAlertView = () => {
 	const onClick2 = onClick(false);
 
 	useEffect( () => {
-		if (window.PalmServiceBridge) {
+		if (window.WebOSServiceBridge ?? window.PalmServiceBridge) {
 			new LS2Request().send({
 				service: 'luna://com.webos.settingsservice/',
 				method: 'getSystemSettings',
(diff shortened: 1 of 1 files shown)


samples/qa-a11y/src/views/ReadAlert.js
```diff
@@ -18,7 +18,7 @@ const ReadAlertView = () => {
 	const onClick2 = onClick(false);
 
 	useEffect( () => {
-		if (window.PalmServiceBridge) {
+		if (window.WebOSServiceBridge ?? window.PalmServiceBridge) {
```
seunghoh, line 21: Do we need this check? 
LS2Request has the check logic and this check logic in this sample looks weird as a reference sample for app developers.

samples/qa-a11y/src/views/ReadAlert.js
```diff
@@ -35,7 +35,7 @@ const ReadAlertView = () => {
 	}, []);
 
 	const onToggle = useCallback(({selected}) => {
-		if (window.PalmServiceBridge) {
+		if (window.WebOSServiceBridge ?? window.PalmServiceBridge) {
```
seunghoh, line 38: Sams as above

seunghoh (approved):
LGTM

samples/qa-a11y/src/views/ReadAlert.js
0x64: Right, `LS2Request` already has checking logic. But, this sample is designed as a sample app, so it looks not weird as an app developer.

PR #1501, opened 2023-09-07
WRP-27297: `Panels.Header` to not show `slotAfter` in incorrect position at the first rendering when `centered` is given

0x64:
### Checklist

When `Panels.Header` is rendered for the first time with `centered` prop, `slotAfter` elements are rendered at unexpected position and they are positioned correctly right after.

### Resolution
To layout `title`, `slotBefore`, and `slotAfter` correctly with the centered title, `slotBefore` and `slotAfter` elements' widths should be measured after rendered. So, `Panels.Header` renders once for measuring and re-renders for the right layout.
In this PR, the logic is changed to hide slots during the inevitable first render for measuring, then to recover slots to show.

### Additional Considerations

### Links
WRP-27297

### Comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact sandstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


seunghoh (approved):
LGTM

PR #1502, opened 2023-09-08
WRP-26318: Fix editable scroller a11y feature to not cut off panel's audio guidance

juwonjeong:
### Checklist

Fix the editable scroller a11y feature to not cut off panel's audio guidance
Audio guide statement updated according to the latest UX

### Resolution
Remove announce() which is used in the focus handler in the editable scroller 
Fix audioGuide to be read with aria-label of iconItem when the iconItem is focused

### Additional Considerations

### Links
WRP-26318

### DIFF
--- a/samples/sampler/stories/qa/IconItem.js
+++ b/samples/sampler/stories/qa/IconItem.js
@@ -261,9 +261,9 @@ export const EditableIcon = (args) => {
 								return (
 									<div key={item.index} className={classNames(css.itemWrapper, {[css.hidden]: item.hidden})} aria-label={`Icon ${item.index}`} data-index={item.index} style={{order: index + 1}} disabled={item.iconItemProps['disabled'] || item.hidden}>
 										<ContainerDivWithLeaveForConfig className={css.removeButtonContainer}>
-											{item.hidden ? null : <Button aria-label="Delete button" className={css.removeButton} onClick={onClickRemoveButton} icon="trash" />}
-											{item.hidden ? null : <Button aria-label="Hide button" className={css.removeButton} onClick={onClickHideButton} icon="minus" />}
-											{item.hidden ? <Button aria-label="Show button" className={css.removeButton} onClick={onClickShowButton} icon="plus" /> : null}
+											{item.hidden ? null : <Button aria-label="Delete" className={css.removeButton} onClick={onClickRemoveButton} icon="trash" />}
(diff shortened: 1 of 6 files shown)


samples/sampler/stories/qa/Panels.js
```diff
@@ -140,3 +153,285 @@ WithAutoFocusControl.parameters = {
...
+	return (
+		<Panels
+			index={panelIndex}
+			noCloseButton
+			onBack={backward}
+		>
+			<Panel>
+				<Header title="Panel with Editable Scroller" subtitle="Click pencil button to start edit mode" />
+				<div ref={divRef}>
+					<Button style={{marginLeft: '36px'}} onClick={forward} icon="edit" />
```
seunghoh, line 347: Let's add `aria-label` to readout "Edit mode"

samples/sampler/stories/qa/IconItem.js
```diff
@@ -260,12 +261,13 @@ export const EditableIcon = (args) => {
 								return (
 									<div key={item.index} className={classNames(css.itemWrapper, {[css.hidden]: item.hidden})} aria-label={`Icon ${item.index}`} data-index={item.index} style={{order: index + 1}} disabled={item.iconItemProps['disabled'] || item.hidden}>
 										<ContainerDivWithLeaveForConfig className={css.removeButtonContainer}>
-											{item.hidden ? null : <Button aria-label="Delete" className={css.removeButton} onClick={onClickRemoveButton} icon="trash" />}
-											{item.hidden ? null : <Button aria-label="Hide" className={css.removeButton} onClick={onClickHideButton} icon="minus" />}
-											{item.hidden ? <Button aria-label="Show" className={css.removeButton} onClick={onClickShowButton} icon="plus" /> : null}
+											{item.hidden ? null : <Button aria-label="Delete button" className={css.removeButton} onClick={onClickRemoveButton} icon="trash" />}
```
seunghoh, line 264: Are you sure? <Button> tag reads "button" so, if we change like this, TTS will reads "Delete button, button"

seunghoh (commented):
These are not related to this PR but, let's follow up on the new ticket
- a11y doesn't tell when Delete/Hide completed
- a11y reads press left/right to move when selecting the hidden button, but the hidden button cannot move. Even, it doesn't need to enter the select mode.
- Long press on the deactivated icon a11y text should not reaout out the edit hint. It should something like "Icon 2 deactivated"

seunghoh (approved):

seunghoh:
LGTM

0x64 (approved):

PR #1512, opened 2023-10-17
WRQ-379: Fix EditableWrapper movement completion announcement

juwonjeong:

### Checklist

Sometimes, focus audio guidance is announced before the movement completion audio guidance.

### Resolution
Fix aria-label of the selected item to be an empty string while reading out movement completion audio guidance

### Additional Considerations

### Links
WRQ-379

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact sandstone module, newest changes on the top.
 
+## [unreleased]
+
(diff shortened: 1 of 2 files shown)


seunghoh (approved):
LGTM as of now but please check below
- Interestingly, long press button edit mode move completion reads the title again only when the storybook mode. full-screen mode doesn't have.
- Please find out root cause why screen reader reads focused item when move completion even focus stays on the same item

PR #1519, opened 2023-11-28
WRQ-1308: Update audio guidance of Dropdown component

juwonjeong:

### Checklist

Update audio guidance of Dropdown component
- When dropdown list is opened, read out `Dropdown list opened`
- When dropdown button is focused, read out `Dropdown` after the contents

### Resolution
Updated aria-label in Dropdown
Updated DropdownList to use aria-labelledby to read only when the focus is entered into the list

### Additional Considerations

### Links
WRQ-1308

### DIFF
--- a/Dropdown/Dropdown.js
+++ b/Dropdown/Dropdown.js
@@ -337,7 +337,7 @@ const DropdownBase = kind({
 			<div {...calcAriaProps} {...rest}>
 				{title}
 				<DropdownButton
-					aria-label={ariaLabel || placeholder + ` ${$L('Dropdown')}`}
+					aria-label={(ariaLabel || placeholder) + ` ${$L('Dropdown')}`}
 					direction={direction}
 					disabled={hasChildren ? disabled : true}
 					focusEffect="static"
--- a/Dropdown/Dropdown.js
+++ b/Dropdown/Dropdown.js
@@ -337,7 +337,7 @@ const DropdownBase = kind({
 			<div {...calcAriaProps} {...rest}>
 				{title}
 				<DropdownButton
-					aria-label={ariaLabel}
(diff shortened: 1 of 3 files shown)


Dropdown/Dropdown.js
```diff
@@ -337,7 +337,7 @@ const DropdownBase = kind({
 			<div {...calcAriaProps} {...rest}>
 				{title}
 				<DropdownButton
-					aria-label={ariaLabel}
+					aria-label={ariaLabel || placeholder + ` ${$L('Dropdown')}`}
```
seunghoh, line 340: You should read "Dropdown Button" when the user sets `aria-label`.

Dropdown/DropdownList.js
```diff
@@ -131,24 +141,27 @@ const DropdownListBase = kind({
...
-				cbScrollTo={scrollTo}
-				dataSize={dataSize}
-				itemSize={ri.scale(itemSize)}
-				style={{
-					height: ri.scaleToRem((itemSize * dataSize) + 36),
-					width: typeof width === 'number' ? ri.scaleToRem(width) : null
-				}}
-			/>
+			<div role="region" aria-labelledby={`${id}_dropdownlist`}>
+				<div id={`${id}_dropdownlist`} aria-label={$L('Dropdown list opened')} />
```
seunghoh, line 153: Good idea!

seunghoh (changes requested):

Dropdown/Dropdown.js
juwonjeong: It's fixed to read "Dropdown Button" even when the user sets `aria-label`.

seunghoh (approved):
LGTM but Let's make a separate ticket for analyzing why it can't read the title when set `aria-label`.

PR #1526, opened 2023-12-04
WRQ-1307: Update audio guidance of TabLayout component 

juwonjeong:
### Checklist

Update audio guidance of TabLayout component

- When the focus is entered into the tabList, read out `{total} items in total {tabname} tab {selectedValue} of {total}`
- When the focus is moved to the other tab, read out `tab {selectedValue} of {total}` after reading the tab name

### Resolution
Update role prop in TabGroup component
Update TabGroup to use aria-labelledby to read only when the focus is entered into the list

### Additional Considerations

### Links
WRQ-1307


seunghoh:
Leave a record for future revisits.
We also fixed double focus issue when pressing the left key on the content area to move to the tab.

seunghoh (changes requested):
Would you please check the scroller size?

seunghoh (approved):
💯

PR #1529, opened 2023-12-11
WRQ-171: Fix `VirtualList` to have proper scroll position when item with affordance is larger than scroll area

0x64:
### Checklist

In a vertical virtual grid list, scroll position for a focused item could be incorrect by 5-way navigation when the scrolling area (client height) is larger than the item height but is smaller than the item height with affordance.

### Resolution
The case was a sort of gray area in a viewpoint of position calculation. Normally layout is based on the item size, but the scroll position is counting also affordance. This condition makes a negative offset on calculation so that an item's top part is cut off.
The intention of the internal calculation allows only 0 or positive offset, so fixed this issue by limiting any negative offset.

### Additional Considerations

### Links
WRQ-171

### Comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The following is a curated list of changes in the Enact sandstone module, newest
 ### Fixed
 
 - `sandstone/Scroller` with `editable` to complete editing when focus left by 5-way key in pointer mode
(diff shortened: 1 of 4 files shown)


juwonjeong (approved):
LGTM

PR #1534, opened 2023-12-26
WRQ-202: Implement editing items in touch mode

juwonjeong:

### Checklist

Implement move/hide/remove items in the editable scroller via touch 

### Resolution
Add touchmove, touchend, dragstart event handler to handle items via touch
Selected items can be moved by dragging the mouse

### Additional Considerations
I think hoverToScroll feature in touch mode needs to be updated
When buttons above the item are touched, it does not highlighted

### Links
WRQ-202

### DIFF
--- a/Scroller/EditableWrapper.js
+++ b/Scroller/EditableWrapper.js
@@ -132,6 +132,7 @@ const EditableWrapper = (props) => {
 
 		lastInputDirection: null,
 
+		isDraggingItem: false,
 		isDragging: false,
 
 		// initialSelected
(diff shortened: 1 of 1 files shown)


juwonjeong:
@0x64 There are few changes according to the review.
- I fixed variable name to `isDragging`.
- I checked if it is possible to make a pair of `handleDragStart` and `handleDragEnd` event handlers. However, it failed. This is because `handleTouchEnd` eventHandler was called before `handleDragEnd` eventHandler. And since editing is finalized in `handleTouchEnd` eventHandler, we can not finalize in `handleDragEnd` eventHandler. We need `handleTouchEnd` event to finalize editing when the item is tapped.

I also added ev.preventDefault() to `handleTouchEnd` event handler to prevent mouse-related events from being called.

0x64:
@juwonjeong Thank you for updating! I got your point. Makes sense.
One issue I found is that there is no way to scroll before selecting. A user may want to scroll to find icon to edit. And, if scrollable, selecting by 'touchend' could lead unexpected selecting after a natural scrolling by drag. So, it also needs to be addressed.

0x64 (approved):

0x64:
@juwonjeong Updated commits look very good! Thank you. 👍

PR #1551, opened 2024-01-22
WRQ-7224: Fixed `Scroller` to stop propagation of 'keyup' event when back key moves focus inside a scroller

0x64:
### Checklist

When a user press back key on a scroll thumb of a scroller with focusableScrollbar that is rendered inside FixedPopupPanels, Scroller moves focus to the scroll body and FixedPopupPanels forwards 'onBack' event to app at the same time. So, the app changes a view to show up another panel. After all, a user feels the scroll body is flashing.

### Resolution
Moving focus to the scroll body is essential in a11y support to read out its contents. Also, there should be a chance to focus the scroll body if there are spottable elements inside the body. To accomplish this goal, 'onBack' event derived from keyup needs to be prevented from dispatching when Scroller catch back key input derived from keydown.
So, fixed `Scroller` to consume a keyup event too when it needs to consume keydown event.

### Additional Considerations

### Links
WRQ-7224

### Comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The following is a curated list of changes in the Enact sandstone module, newest
 ### Fixed
 
 - `sandstone/Scroller` with `editable` prop to remain focused on the selected item when completing edit by down or enter key in pointer mode
+- `sandstone/Scroller` not to forward `onBack` handler of `sandstone/Panels` when focus moves from scroll thumb to BodyText via back key
 - `sandstone/VirtualList` to have proper scroll position when item with affordance is larger than scroll area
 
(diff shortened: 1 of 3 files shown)


seunghoh:
Let's add the unit test to cover the updated functions. 
Don't we need to add this scenario on qa-sampler?

0x64:
Note for merger: to avoid an expected conflict of CHANGELOG.md, I did not update it. Before merging, please let me know so that I can commit the changelog.

juwonjeong (approved):
LGTM

seunghoh (approved):
LGTM

PR #1557, opened 2024-02-06
WRQ-6188: Added FloatingPopup prototype

mmyelyn:
### Checklist

There is a requirement for a floating popup capable of free positioning.
This should also be able to receive content such as images and text.
This is a prototype implementation of the requirements.

### Resolution
The component provides a container so that the app can pass contents inside the popup.
The component allows you to specify the desired location.
In default sampler, we guide you on how to use it when passing images and text in the sampler.

### Additional Considerations
Coordinate conversion may be required as the screen changes.
RTL may be required.
may need to trigger click events on components that are not hidden by the container.

### Links
WRQ-6188

### Comments
Enact-DCO-1.0-Signed-off-by: Hyelyn Kim ([myelyn.kim@lge.com](mailto:sc.baek@lge.com))

### DIFF
--- a/FloatingPopup/FloatingPopup.js
+++ b/FloatingPopup/FloatingPopup.js
@@ -1,0 +1,162 @@
+/**
+ * Modal component available with free positioning.
+ *
+ * @example
+ * <FloatingPopup open position={left: 50, top:50}>
+ *   <FloatingPopupContainer>
+ *     {* App Contents *}
+ *   <FloatingPopupContainer/>
+ * </FloatingPopup>
+ *
+ * @module sandstone/FloatingPopup
+ * @exports FloatingPopup
+ * @exports FloatingPopupBase
+ */
+
+import FloatingLayer from '@enact/ui/FloatingLayer';
+import kind from '@enact/core/kind';
+import {Component} from 'react';
+import PropTypes from 'prop-types';
+import {forward, forwardCustom} from '@enact/core/handle';
+
+import css from './FloatingPopup.module.less';
+
+/**
+ * The Floating Popup base for app contents.
+ *
+ * @class FloatingPopupBase
+ * @memberof sandstone/FloatingPopup
+ * @ui
+ * @public
+ */
+const FloatingPopupBase = (props) => {
+	return (
+		<>{props.children}</>
+	);
+};
+
+/**
+ * An item for FloatingPopup.
+ *
+ * @class FloatingPopupContainer
+ * @memberof sandstone/FloatingPopup
+ * @ui
+ * @public
+ */
+const FloatingPopupContainer = kind({
+	name: 'FloatingPopupContainer',
+	propTypes: /** @lends sandstone/FloatingPopup.FloatingPopupContainer.prototype */ {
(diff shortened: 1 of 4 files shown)


seunghoh:
I have some questions about the props and component naming.
- Popup component already renders in a FloatingLayer. FloatingPopup sounds confusing to me. What do you think?
- Don't we need a prop for supporting on/off RTL positioning? When I change the locale to the ar-SA, icon and text flips which makes sense but the position remains. Some cases may remain but some cases may change the position to the right-based align. What do you think?

mmyelyn:
@seunghoh 
> I have some questions about the props and component naming.
> 
> * Popup component already renders in a FloatingLayer. FloatingPopup sounds confusing to me. What do you think?
> * Don't we need a prop for supporting on/off RTL positioning? When I change the locale to the ar-SA, icon and text flips which makes sense but the position remains. Some cases may remain but some cases may change the position to the right-based align. What do you think?

1. I agree. So I changed naming to PopupContainer from FloatingPopup. This means that it is a container with the function of creating a popup.

2. Good. I added a prop for supporting on/off RTL position. But I should check whether it is needed under the normal circumstances of the app through the guide. If necessary, i should check whether the app wants to change only the location or the content direction as well.

PopupContainer/PopupContainer.module.less
```diff
@@ -0,0 +1,9 @@
+// PopupContainer.module.less
+
+.PopupContainer {
+    .container {
+        position: absolute;
+        min-height: 150px;
```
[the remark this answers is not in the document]
mmyelyn: I don't think it's necessary. This was added when it was not implemented as a container. Removed.

PopupContainer/PopupContainer.js
```diff
@@ -0,0 +1,206 @@
...
+	handleClose = (ev) => {
+		forwardCustom('onClose')(ev, this.props);
+	};
+
+	handleOpen = (ev) => {
+		forward('onOpen', ev, this.props);
+	};
+
+	render () {
+		const {className, open, position, children, forceDirection, ...rest} = this.props;
```
0x64, line 185: `position` and `forceDirection` can be passed to `PopupContainerBase` via `rest` if they are destructed here.

mmyelyn: Does this mean it should be deleted from this line? I want to explicitly show that `PopupContainerBase` should receive those props. what do you think?

0x64: Some components in our libraries get a lot of props and some components just get one or two. So, we can't expect every component has intentionally exposed props in render block. That means, developers should not rely on exposed props. So, I believe we don't need to keep it as we cannot rely on it.

mmyelyn: I like it to be explicit for readability, but I agree with you. Currently, component unity seems more important. I cleaned up the properties.

0x64 (approved):

0x64:
LGTM. We don't have a fixed requirement yet, so I will keep this as a draft PR.

PR #1566, opened 2024-02-20
WRQ-6220: Fixed `VirtualList` to show the focused item properly when switching from pointer mode to 5-way mode

0x64:
### Checklist

If there is no navigable target when switching from pointer mode to 5-way mode on a partially visible focused item, the focus remains on the focused item, but the item still does not show its whole content as there is no scrolling.

### Resolution
Fixed to re-focus only when no focus moving by switching from pointer mode to 5-way mode, so that a scrolling occurs.

### Additional Considerations

### Links
WRQ-6220

### Comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact sandstone module, newest changes on the top.
 
+## [unreleased]
+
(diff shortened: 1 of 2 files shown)


juwonjeong (approved):
LGTM

PR #1578, opened 2024-03-08
WRQ-4796: Add comment to Dropdown qa-a11y sample

juwonjeong:

### Checklist

In qa-a11y Dropdown sample, there are samples that the aria-label and title are both set. We add a guide on how to read out these samples. 

### Resolution
Add comment to the qa-a11y dropdown sample that when aria-label and title are both given, only aria-label of dropdown will be read out

### Additional Considerations

### Links
WRQ-4796


seunghoh (approved):
LGTM

PR #1583, opened 2024-04-01
WRQ-15475: Implement A11y for PageViews

juwonjeong:

### Checklist

A11y for PageViews needs to be supported.

### Resolution
Implement read-out feature for the PageViews component with using `region` and `aria-labelledby`.

### Additional Considerations

### Links
WRQ-15475


PageViews/PageViews.js
```diff
@@ -37,6 +38,7 @@ import css from './PageViews.module.less';
 const PageViewsBase = kind({
 	name: 'PageViews',
 	propTypes: /** @lends sandstone/PageViews.PageViewsBase.prototype */ {
+		'aria-label': PropTypes.string,
```
seunghoh, line 41: Let's update the example with `aria-label`
and add one blank line :)

seunghoh (changes requested):

PageViews/PageViews.js
juwonjeong: I updated example with `aria-label` and added blank line.

seunghoh (approved):

seunghoh:
LGTM

PR #1602, opened 2024-04-26
WRQ-9559: Fix scrolling by holding pageup/down key

vJIYEv:
### Checklist

Scrolling by holding pageup/down key works / does not work case by case.
- Pointer mode
  - holding pageup/down key works if pointer is out of the component.
  - holding pageup/down key does not work if pointer is in the component.
- Key mode
  - holding pageup/down key does not work in macOS WebKIt (Safari). (It works in other environments)

We need to fix for all cases to show same behavior.

### Resolution
Scrolling by holding pageup/down key occurs because of default behavior of browser.
But default behavior varies from browser to browser, so we can't make scroll works if default behavior is not to scroll.
To match scroll behavior of all cases, I prevented default behavior when pressed pageup/down key.

### Additional Considerations

### Links
WRQ-9559

### Comments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ dist: focal
 language: node_js
 node_js:
     - lts/*
-    - node
+    - 21
 sudo: false
 before_install:
     - sudo apt-get update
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -2,7 +2,7 @@ dist: focal
 language: node_js
 node_js:
     - lts/*
-    - node
+    - 21
```
0x64, line 5: Let's revert it back. :)

PR #1604, opened 2024-04-30
WRQ-15921: Fix `Popup` to not focus the element of the popup container when popup is hidden

juwonjeong:
### Checklist

In FixedPopupPanels when popup is closed, focus does not restore into the background component. It restore focus to the element in the popup panels.

### Resolution
If the FixedPopupPanels is open when the sample is first rendered, the activator is set to null.

In this case, to restore the focus spotlight module use `getTargetByContainer()`.
In `getTargetByContainer()`, they use `getContainerToSearch()` to get a list of containers to search the focus target, and call `getContainerFocusTarget()` for each container in this list to find the element to focus on. 

Since the popup container is the last container, it is included in the search list and even has the highest priority.

So to fix this issue, we need to prevent popup container to be included in the result of the `getContainerToSearch()`
function. So I added logic to filter popup container with disabling container's selector. 

I fixed this issue by disabling lastContainerId when activator is null.
### Additional Considerations

### Links
WRQ-15921


0x64:
The bug occurs when `activator` is not correctly set. So, I prefer to show up the issue condition directly. I'm not sure that what we're supposed to do when `current` does not exist, but it looks okay to handle this case with the same manner. My idea looks like the below. (Just idea, not tested fully.)

>		if (!current || (containerNode && containerNode.contains(current))) {
>			// attempt to set focus to the activator, if available
>			if (!Spotlight.isPaused()) {
>				if (activator) {
>					if (!Spotlight.focus(activator)) {
>						Spotlight.focus();
>					}
>				} else {
>					Spotlight.disableSelector(lastContainerId);
>					Spotlight.focus();
>					Spotlight.enableSelector(lastContainerId);
>				}
>			}
>		}

0x64:
The bug occurs when `activator` is not correctly set. So, I prefer to show up the issue condition directly.

>		if (!current || (containerNode && containerNode.contains(current))) {
>			// attempt to set focus to the activator, if available
>			if (!Spotlight.isPaused()) {
>				if (activator) {
>					if (!Spotlight.focus(activator)) {
>						Spotlight.focus();
>					}
>				} else {
>					Spotlight.disableSelector(lastContainerId);
>					Spotlight.focus();
>					Spotlight.enableSelector(lastContainerId);
>				}
>			}
>		}

juwonjeong:
> The bug occurs when `activator` is not correctly set. So, I prefer to show up the issue condition directly. I'm not sure that what we're supposed to do when `current` does not exist, but it looks okay to handle this case with the same manner. My idea looks like the below. (Just idea, not tested fully.)
> 
> > ```
> >   if (!current || (containerNode && containerNode.contains(current))) {
> >   	// attempt to set focus to the activator, if available
> >   	if (!Spotlight.isPaused()) {
> >   		if (activator) {
> >   			if (!Spotlight.focus(activator)) {
> >   				Spotlight.focus();
> >   			}
> >   		} else {
> >   			Spotlight.disableSelector(lastContainerId);
> >   			Spotlight.focus();
> >   			Spotlight.enableSelector(lastContainerId);
> >   		}
> >   	}
> >   }
> > ```

Thank you for the review. I agree with your comment that activator should be used as a condition. I tested it and it seems to work fine.

0x64 (approved):
Looks good to me!

Popup/Popup.js
```diff
@@ -611,15 +612,24 @@ class Popup extends Component {
 
 		const current = Spotlight.getCurrent();
 		const containerNode = getContainerNode(this.state.containerId);
+		const lastContainerId = getLastContainer();
```
MikyungKim, line 615: Do we need to know the last container id? Or can we use the containerId of this Popup(`this.state.containerId`)?

tests/ui/specs/FixedPopupPanels/WithPopupOpen/FixedPopupPanelsWithPopupOpen-specs.js
```diff
@@ -0,0 +1,21 @@
...
+
+describe('FixedPopupPanels', function () {
+	const Interface = Page.fixedPopupPanels;
+
+	beforeEach(async function () {
+		await Page.open('WithPopupOpen');
+	});
+
+	describe('5-way', function () {
+
```
MikyungKim, line 11:
```suggestion
```

tests/ui/specs/FixedPopupPanels/WithPopupOpen/FixedPopupPanelsWithPopupOpen-specs.js
```diff
@@ -0,0 +1,21 @@
+const Page = require('./FixedPopupPanelsWithPopupOpenPage');
+
+describe('FixedPopupPanels', function () {
```
MikyungKim, line 3:
```suggestion
describe('FixedPopupPanelsWithPopupOpen', function () {
```

Popup/Popup.js
juwonjeong: As I checked, we need to know the last container id.  For example, in the sample code lastContainerId is `panel-container5` and this.state.containerId returns `container-2`. So when we disable this.state.containerId(=`container-2`), spotlight logic still call `getContainerFocusTarget()` with `panel-container5` which will reproduce the issue.

tests/ui/specs/FixedPopupPanels/WithPopupOpen/FixedPopupPanelsWithPopupOpen-specs.js
juwonjeong: It's fixed.

tests/ui/specs/FixedPopupPanels/WithPopupOpen/FixedPopupPanelsWithPopupOpen-specs.js
juwonjeong: It's fixed.

MikyungKim (approved):
LGTM

PR #1629, opened 2024-06-19
WRQ-25442: Fixed Editable Scroller to blur the focused item properly

MikyungKim:
### Checklist

In edit mode, the focused item in Editable Scroller seems not properly blurred when the pointer leaves the item.

### Resolution
The problem here is that there is `focusItemRef` to focus the IconItem by pointer but there is no way to blur it when the pointer leaves. Since we are manually adding `.focused` class by function.
So I added `blurItemRef` to get the function to blur the item and call it when the pointer leaves the item container.
If I call the function when the IconItem gets blurred, there is no way to select the additional button showing above (e.g. remove button).

### Additional Considerations
I found the remove buttons of focused item remained when pressing 'up' key from the pointer resides in the empty space like below. So I added the condition to `handleGlobalKeyDownCapture` to make sure those icons are not showing.
image
This PR is the same modification with https://github.com/enactjs/sandstone/pull/1628 based on `develop` branch.

### Links
WRQ-25442

### Comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact sandstone module, newest changes on the top.
 
+## [unreleased]
+
+### Added
+
+- `sandstone/Scroller` `editable.blurItemFuncRef` prop to provide a function for blurring the focused item
+
 ## [2.9.0-beta.1] - 2024-06-17
 
 ### Added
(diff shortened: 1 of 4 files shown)


seunghoh (commented):
For the additional consideration case, you fixed to remove focused effect on IconItem but still I see the trash icon.

juwonjeong (approved):
LGTM

PR #1631, opened 2024-06-20
WRQ-24170: Fixed VirtualGridList qa sampler by passing args as a prop

vJIYEv:
### Checklist

In qa-sampler > VirtualGridList > `with Button, Spotlight goes to correct target`, scroll with wheel did not work.

### Resolution
`scrollMode` value of VirtualGridList in the sample was `undefined` and this caused the issue.
I added control of `scrollMode`, which have `native` and `translate` value, so `scrollMode` value is passed into VirtualGridList and wheel works well.

### Additional Considerations

### Links
WRQ-24170

### Comments

### DIFF
--- a/samples/sampler/stories/qa/VirtualGridList.js
+++ b/samples/sampler/stories/qa/VirtualGridList.js
@@ -119,13 +119,17 @@ class MyVirtualList extends Component {
 					direction="vertical"
 					itemRenderer={this.renderItem}
 					itemSize={{minWidth: ri.scale(570), minHeight: ri.scale(156)}}
-					key={select('scrollMode', prop.scrollModeOption, Config)}
-					scrollMode={select('scrollMode', prop.scrollModeOption, Config)}
+					key={this.props.args['scrollMode']}
+					scrollMode={this.props.args['scrollMode']}
 				/>
 			</div>
 		);
 	}
 }
+
+MyVirtualList.propTypes = {
+	args: PropTypes.array
+};
 
 class ButtonAndVirtualGridList extends Component {
 	constructor (props) {
@@ -136,7 +140,7 @@ class ButtonAndVirtualGridList extends Component {
 	}
 
 	renderPopup = (rest) => {
-		return <MyVirtualList {...rest} closePopup={this.closePopup} />;
+		return <MyVirtualList {...rest} closePopup={this.closePopup} args={this.props.args} />;
 	};
 
 	openPopup = () => {
@@ -166,6 +170,7 @@ class ButtonAndVirtualGridList extends Component {
 }
 
 ButtonAndVirtualGridList.propTypes = {
+	args: PropTypes.array,
 	rtl: PropTypes.bool
 };
 
@@ -221,7 +226,7 @@ HorizontalVirtualGridList.parameters = {
 	propTables: [Config]
 };
 
-export const WithButtonSpotlightGoesToCorrectTarget = () => <ButtonAndVirtualGridListSamples />;
+export const WithButtonSpotlightGoesToCorrectTarget = (args) => <ButtonAndVirtualGridListSamples args={args} />;
 
 WithButtonSpotlightGoesToCorrectTarget.storyName = 'with Button, Spotlight goes to correct target';
 WithButtonSpotlightGoesToCorrectTarget.parameters = {
@@ -229,6 +234,8 @@ WithButtonSpotlightGoesToCorrectTarget.parameters = {
 		hideNoControlsWarning: true
 	}
 };
+
+select('scrollMode', WithButtonSpotlightGoesToCorrectTarget, prop.scrollModeOption, Config);
 
 export const HorizontalSquaredVirtualGridList = (args) => (
 	<VirtualGridList
--- a/samples/sampler/stories/qa/VirtualGridList.js
+++ b/samples/sampler/stories/qa/VirtualGridList.js
@@ -134,14 +134,13 @@ MyVirtualList.propTypes = {
 class ButtonAndVirtualGridList extends Component {
 	constructor (props) {
 		super(props);
-		this.scrollMode = props.scrollMode;
 		this.state = {
 			isPopup: false
 		};
 	}
 
 	renderPopup = (rest) => {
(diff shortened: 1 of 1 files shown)


seunghoh:
You missed checking `At least one test case is included for this feature or bug fix`

seunghoh (changes requested):
1. Why do you need to add `scrollMode` into the control? 
2. Why default value of the `scrollMode` is `translate`

vJIYEv:
@seunghoh
1. Because we had `scrollMode` control in this sample before.
2. As I checked, default value is `native`, not `translate`.

MikyungKim:
Actually the `scrollMode` is `undefined` that was the root cause of this issue.

seunghoh:
@vJIYEv Please update the PR description with more details including root cause.

seunghoh (approved):

samples/sampler/stories/qa/VirtualGridList.js
```diff
@@ -136,7 +140,7 @@ class ButtonAndVirtualGridList extends Component {
 	}
 
 	renderPopup = (rest) => {
-		return <MyVirtualList {...rest} closePopup={this.closePopup} />;
+		return <MyVirtualList {...rest} closePopup={this.closePopup} args={this.props.args} />;
```
MikyungKim, line 143:
```suggestion
		return <MyVirtualList {...rest} args={this.props.args} closePopup={this.closePopup} />;
```
You can do like this but if you want to pass some props to `popupComponent`, you can also use `popupProps` prop.

samples/sampler/stories/qa/VirtualGridList.js
```diff
@@ -119,14 +119,18 @@ class MyVirtualList extends Component {
 					direction="vertical"
 					itemRenderer={this.renderItem}
 					itemSize={{minWidth: ri.scale(570), minHeight: ri.scale(156)}}
-					key={select('scrollMode', prop.scrollModeOption, Config)}
-					scrollMode={select('scrollMode', prop.scrollModeOption, Config)}
+					key={this.props.args['scrollMode']}
+					scrollMode={this.props.args['scrollMode']}
```
MikyungKim, line 123: You don't need to refer `this` since we already have `props` in render function.
```suggestion
					key={props.args['scrollMode']}
					scrollMode={props.args['scrollMode']}
```

MikyungKim (changes requested):
Have you checked console log when you are running your sample?
I'm seeing warning message saying like below. How can we fix this problem?
error

Also, when I'm inspecting elements I'm seeing `args` attribute in `div`. I'm sure you don't want this in `div`. How can we fix this one?
args

Don't forget to check console log for runtime warnings or errors next time.

samples/sampler/stories/qa/VirtualGridList.js
vJIYEv: I choosed this way following how `closePopup` is passed.

samples/sampler/stories/qa/VirtualGridList.js
vJIYEv: OK

samples/sampler/stories/qa/VirtualGridList.js
```diff
@@ -84,6 +84,11 @@ const ContextualPopupButton = ContextualPopupDecorator(Button);
 let lastIndex = 0;
 
 class MyVirtualList extends Component {
+	constructor (props) {
+		super();
+		this.scrollMode = props.scrollMode;
```
MikyungKim, line 89: Why do we need to do this?

vJIYEv: In line 120, `props` are passed to `<div>` elem but we don't need to pass `props.scrollMode` to this `<div>`. (same problem with `args="[object Object]"` you saw in inspector)
So I deleted `props.scrollMode` in line 117.
But we need `scrollMode` value in `VirtualGridList`, so I took `scrollMode` value from `props` before deleting it.
That's why we need this code.

MikyungKim: :) I know what we need to do but the way you handled is not what I expected.
Could you try to use [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) instead of this?
If you search in our code, you can find many usages.

vJIYEv: OK, I changed this to use destructuring assignment.

samples/sampler/stories/qa/VirtualGridList.js
```diff
@@ -119,24 +119,29 @@ class MyVirtualList extends Component {
...
 }
 
+MyVirtualList.propTypes = {
+	scrollMode: PropTypes.string
+};
+
 class ButtonAndVirtualGridList extends Component {
 	constructor (props) {
 		super(props);
+		this.scrollMode = props.scrollMode;
```
MikyungKim, line 137: Please fix here too.

vJIYEv: OK

samples/sampler/stories/qa/VirtualGridList.js
```diff
@@ -108,7 +108,7 @@ class MyVirtualList extends Component {
 	};
 
 	render () {
-		let props = {...this.props};
+		let {scrollMode, ...props} = this.props;
```
MikyungKim, line 111: It's ok but we usually name the rest props to `rest`.

```suggestion
		const {scrollMode, ...rest} = this.props;
```

vJIYEv: OK

samples/sampler/stories/qa/VirtualGridList.js
```diff
@@ -230,6 +235,8 @@ WithButtonSpotlightGoesToCorrectTarget.parameters = {
 	}
 };
 
+select('scrollMode', WithButtonSpotlightGoesToCorrectTarget, prop.scrollModeOption, Config);
```
MikyungKim, line 238: Let's move this line to 231 before the definition of storyName as other sample does.

MikyungKim (approved):
LGTM

PR #1643, opened 2024-07-18
WRQ-16552: Delete controls that cannot be tested in the `Header` of qa-sampler

mmyelyn:
### Checklist

There is an issue where the back button is not visible even if noBackButton is set to false in the Header component of qa-sampler.
After looking up the history, I found that this was the correct behavior. 

In the header, even if noBackButton is false, the back button will not appear if backButtonAvailable is false. 
At this time, backButtonAvailable checks the index and type of the panel. In the case of the first panel with index 0, the back button is not exposed. 

Therefore, in the current implementation where only the first panel exists, the back button does not appear, so noBackButton and backButtonAvailable controls are not needed.

Please see: https://github.com/enactjs/sandstone/pull/211, https://github.com/enactjs/sandstone/pull/819

### Resolution
Removed noBackButton and backButtonAvailable controls, which cannot be tested in the current implementation.

### Additional Considerations
If the sample is modified to meet the conditions, the noBackButton control should be restored.

### Links
WRQ-16552

### Comments


MikyungKim (approved):
LGTM

PR #1657, opened 2024-08-07
WRR-977: Fixed `Scroller` with `editable` prop to move items or complete edit mode when keydown event fired in pointer mode

juwonjeong:
### Checklist

When selecting an item in an editable scroller and pressing the 5-way keys (e.g. Left, Right, Enter, Down) in pointer mode, the item does not move or the edit mode does not complete.

### Resolution
When 5-way key is pressed in pointer mode, only `handleGlobalkeyDownCapture` function is called. So I added logic in global keydown handler. 

### Additional Considerations
* In qa-sampler, when exiting edit mode after pressing a key multiple times, a blurItem event sometimes occurs. Then no item will be in focus. So I fixed to check the pointer mode in blurItem handler. 
240807_blurItem

* Removed CSS which makes focus effect of qa-sampler inconsistent in this PR. 
** The Pink item does not scale even though it get focused by key because the mouse is hovering over the removeContainer of the pink item.  As it was not the UX guide and was only implemented in the sampler, I removed it to avoid confusion later. 
240806_hover

* I fixed the additional issue with 87531b99885382ae69b9e5b18be802bc7cd99f52 commit.
** When button is focused with mouse and press the left/right button, it should move the focus instead of moving the item.

### Links
WRQ-30470


seunghoh:
For the 2nd case of the additional considerations, it won't be an issue because
1. The Pointer will be removed in 5way mode on webOS devices
2. On PC, it may occur but too edge case (holder the mouse and move focus with the keyboard)

Scroller/EditableWrapper.js
```diff
@@ -262,7 +262,7 @@ const EditableWrapper = (props) => {
 
 	const blurItem = useCallback((target) => {
 		const itemNode = findItemNode(target);
-		if (itemNode && !mutableRef.current.selectedItem) {
+		if (getPointerMode() && itemNode && !mutableRef.current.selectedItem) {
```
MikyungKim, line 265: Let's handle this from app side.

Scroller/EditableWrapper.js
```diff
@@ -697,10 +699,23 @@ const EditableWrapper = (props) => {
...
+					completeEditingByKeyDown();
+				} else if (is('up', keyCode) && nextTarget.getAttribute('role') !== 'button') {
+					setPointerMode(false);
+					Spotlight.focus(nextTarget);
+				}
+			} else if (is('enter', keyCode)) {
+				if (selectedItem) {
+					completeEditingByKeyDown();
+				} else {
+					startEditing(focusedItem);
```
MikyungKim, line 714: How about
```suggestion
				} else
					handleKeyDownCapture(ev);
				}
			} else if (is('enter', keyCode)) {
				handleKeyDownCapture(ev);
```

Scroller/EditableWrapper.js
juwonjeong: It's fixed.

Scroller/EditableWrapper.js
juwonjeong: After testing, I found that the `handleKeyDownCapture` function does not work as we expect because the ev.target is the `body` element, not each item. Therefore, it seems difficult to change it to use `handleKeyDownCapture`.

MikyungKim: T.T

MikyungKim: I'll add some inline comments.

juwonjeong:
@MikyungKim, I tried to add some comments on the added logic. Thank you.

Scroller/EditableWrapper.js
```diff
@@ -687,20 +687,43 @@ const EditableWrapper = (props) => {
...
 			if (direction) {
 				const nextTarget = getTargetByDirectionFromPosition(direction, position, mutableRef.current.spotlightId);
 
 				if (!scrollContainerRef.current.contains(nextTarget)) {
+					// If the nextTarget is not contained in scrollContainerRef, complete editing
 					const orders = finalizeOrders();
 					finalizeEditing(orders);
+				} else if ((is('left', keyCode) || is('right', keyCode)) && selectedItem) {
+					// When an item is selected and press the `left` or `right` key, move the selectedItem in that direction
+					moveItemsByKeyDown(ev);
```
MikyungKim, line 707: As we discussed, the focus seems to move after this. Let's prevent the moving focus.

juwonjeong: It's fixed.

MikyungKim:
For the record, to reproduce the issue resolved by https://github.com/enactjs/sandstone/pull/1657/commits/87531b99885382ae69b9e5b18be802bc7cd99f52, do the followings.
1. Enter editing mode by pointer
2. Click on an item to start editing
3. Press Left key while the pointer is between the space upper items and icon item like below.
image
4. Make sure the spotlight is on the upper icon 
image
5. Move the pointer little bit so the icon item returned its originated position.
6. Press left or right key
Expected result: The item will be moved.
Actual result: The key input doesn't move the item.

Scroller/EditableWrapper.js
```diff
@@ -623,7 +623,7 @@ const EditableWrapper = (props) => {
 			mutableRef.current.needToPreventEvent = true;
 		} else if (is('left', keyCode) || is('right', keyCode)) {
 			if (selectedItem) {
-				if (mutableRef.current.lastKeyEventTargetElement?.getAttribute('role') !== 'button') {
+				if (mutableRef.current.lastKeyEventTargetElement?.getAttribute('role') !== 'button' && target.getAttribute('role') !== 'button') {
```
MikyungKim, line 626: I found one another issue seems to be caused by `lastKeyEventTargetElement` related condition.
Reproduciable Step
1. Start editing by clicking on an item
2. Move the spotlight to the upper button
3. Press left or right key
4. Move the cursor to the icon item
5. Press left or right key
Expected result: The item moves left or right
Actual result: The item doesn't move

I think we could remove the fore condition.
And as we discussed, let's dicuss with UX team for the clearer scenario for this.
e.g. How many buttons can be in the upper area? How should spotlight move from the last button when we press right key, etc.

juwonjeong: It's fixed.

Scroller/EditableWrapper.js
```diff
@@ -687,20 +682,45 @@ const EditableWrapper = (props) => {
 	}, [completeEditingByKeyDown]);
 
 	const handleGlobalKeyDownCapture = useCallback((ev) => {
-		if (getPointerMode() && !scrollContainerRef.current.contains(Spotlight.getCurrent()) && (mutableRef.current.selectedItem || mutableRef.current.focusedItem)) {
+		const {focusedItem, selectedItem} = mutableRef.current;
+
+		// If the pointer mode is `true` and the focused component is not contained in scrollContainerRef,
+		// only `handleGlobalKeyDownCapture` is called instead of `handleKeyDownCapture`
```
MikyungKim, line 688: It'd be better to add
```suggestion
		// only `handleGlobalKeyDownCapture` is called instead of `handleKeyDownCapture`
		// Below is mainly for handling key pressed while pointer mode is `true`.
```

juwonjeong: It's fixed.

CHANGELOG.md
```diff
@@ -18,6 +18,10 @@ The following is a curated list of changes in the Enact sandstone module, newest
 - `sandstone/Scroller`, `sandstone/Slider`, and `sandstone/VirtualList` to have default prop when `undefined` prop is passed
 - `sandstone/Steps` prop `size` to accept number type
 
+### Fixed
```
MikyungKim, line 21: I think this one went wrong.

juwonjeong: Sorry, It's fixed.

samples/sampler/stories/qa/Scroller.js
```diff
@@ -11,6 +11,7 @@ import {ContentContainerDecorator, Scroller} from '@enact/sandstone/Scroller';
 import {VirtualGridList} from '@enact/sandstone/VirtualList';
 import Spotlight from '@enact/spotlight';
 import SpotlightContainerDecorator from '@enact/spotlight/SpotlightContainerDecorator';
+import {getPointerMode} from '@enact/spotlight/src/pointer';
```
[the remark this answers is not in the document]
juwonjeong: It's fixed.

MikyungKim (approved):
LGTM

0x64 (approved):

PR #1576, opened 2024-03-07
WRQ-6498: Remove LS2Request call from theme library

adrian-cocoara-lgp:
### Checklist

* [x] I have read and understand the [contribution guide](http://enactjs.com/docs/developer-guide/contributing/)
* [ ] A [CHANGELOG entry](http://enactjs.com/docs/developer-guide/contributing/changelogs/) is included
* [ ] At least one test case is included for this feature or bug fix
* [x] Documentation was added or is not needed
* [ ] This is an API breaking change

### Issue Resolved / Feature Added
[//]: # (Describe the issue resolved or feature added by this pull request)
Removed LS2Request call from `ThemeDecorator`

### Resolution
[//]: # (Does the code work as intended?)
[//]: # (What is the impact of this change and *why* was it made?)
Extracted LS2Request call logic from `ThemeDecorator` and added it under `enact/webos/lastInputType`
Now `sandstone/ThemeDecorator` calls the new method instead of calling LS2Request directly

### Additional Considerations
[//]: # (How should the change be tested?)
[//]: # (Are there any outstanding questions?)
[//]: # (Were any side-effects caused by the change?)

### Links
[//]: # (Related issues, references)
WRq-6498

### Comments
Enact-DCO-1.0-Signed-off-by: Adrian Cocoara adrian.cocoara@lgepartner.com

PR #1687, opened 2024-09-12
WRR-1010: `Scroller`: Fixed to focus content area properly when `focusableScrollbar` prop is `byEnter`

0x64:
### Checklist

The last input type is not recognized properly so that Scroller does not focus the contents well when `focusableScrollbar` is `byEnter`.

### Resolution
The issue occurs after #1576. When the new Enact API is added and Sandstone is changed to use it, the logic to check if the platform is supported became unclear. Now, we decided to let Enact to simply call webOS specific API without checking and to make Sandstone to check the availability before call the API since the feature it handles is just for specific platforms that Sandstone determines.

### Additional Considerations

### Links
WRR-1010
enactjs/enact#3278
Related: WRQ-6498 / #1576

### Comments


MikyungKim (approved):
LGTM

juwonjeong (approved):
LGTM

juwonjeong:
codecov is decreased because test case is removed from ThemeDecorator-specs.js

PR #1699, opened 2024-09-25
WRR-4572: Fix `Scroller` to read out its content properly when it is located in `Panel`

juwonjeong:

### Checklist

When a `Scroller` with a `focusableScrollbar` contains a button and is located in a `Panel`, the read-out feature behaved differently than we expected.
Screen reader read "Button A" before "This is a panel" was fully read, so the user would only hear the aria label of the `button`.
```
<Panel aria-label = 'This is a panel'>
    <Header title = 'Title' />
        <Scroller focusableScrollbar>
            <Button onClick={() => this.setState({index: 0})}>Button A</Button>
        </Scroller>
</Panel>
```

### Resolution
* This issue occurred because when `focusableScrollbar` was enabled, Spotlight first focused the SpotlightPlaceholder and then focused the scroller's content. This caused the screen reader to read the panel's aria-label and the button's aria-label separately.
* So I added `aria-hidden` attribute to `ScrollbarPlaceholder` to removed it from the Accessibility tree.

### Additional Considerations

### Links
WRR-4572


seunghoh (approved):
LGTM

PR #1702, opened 2024-09-26
WRR-4590: Fix to focus the topmost element after scroll by voice control

0x64:
### Checklist

When a focusable node inside a scroller is sticked to the top, it cannot be focused after scroll by voice control.

### Resolution
The cause is a logic to check if the candidate focus target is inside of a scroller viewport. When a focusable node is touching the boundary of a viewport, the existing logic does not consider the node as visible since it missed simply '='.

I updated to recognize the case that boundaries are the same.

### Additional Considerations

I also refine some code with another commit.

### Links
WRR-4590

### Comments


seunghoh (approved):
LGTM

PR #1704, opened 2024-09-26
WRR-4591: Fix `VideoPlayer` to show only the mini feedback when pressing play/pause key

mmyelyn:
### Checklist

* [X] A [CHANGELOG entry](http://enactjs.com/docs/developer-guide/contributing/changelogs/) is included

There is an issue where the media slider appears when you press the play/pause button on the remote control.
But only mini feedback content should appear.

### Resolution
I set the initial value for playbackRate.

Previously, the playbackRate value was always set via the setPlaybackRate function when the play and pause functions were called.
Now, playbackRate is set only when in seeking mode when the play/pause functions were called. Therefore, the initial value of playbackRate is missing and set to undefined. Because of this, the visibility conditions of the media slider have changed.

### Additional Considerations

### Links
WRR-4591

### Comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact sandstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `sandstone/VideoPlayer` to show only the mini feedback using play/pause key
+
 ## [2.9.2] - 2024-09-26
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact sandstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `sandstone/VideoPlayer` to show only the mini feedback using play/pause key
```
```suggestion
- `sandstone/VideoPlayer` to show only the mini feedback when pressing play/pause key
```

MikyungKim (approved):
LGTM

PR #1707, opened 2024-09-30
WRR-4660: Fixed `PageViews` to not clip the shadow of navigation buttons

juwonjeong:

### Checklist

The issue is that the navButtonContainer area of the PageViews was too tight, so the shadow of navigation buttons was clip when `fullContents` prop is `true`.

### Resolution
Set the height of the `navButtonContainer` to twice the height of the button.
And the formula related to `top`is a calculation that will position the navButtonContainer in the center.

### Additional Considerations

### Links
WRR-4660

### DIFF
--- a/PageViews/PageViews.module.less
+++ b/PageViews/PageViews.module.less
@@ -19,10 +19,9 @@
 	.navButtonContainer {
 		position: absolute;
 		width: 100%;
-		height: @sand-button-height * 2;
-		top: ~"calc(50% - " (@sand-button-height * 2) / 2 ~")";
+		height: @sand-button-height;
+		top: ~"calc(50% - "@sand-button-height / 2 ~")";
 		z-index: 10;
-		overflow: hidden;
 	}
 
 	&.number .navButton {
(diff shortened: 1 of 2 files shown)


PageViews/PageViews.module.less
```diff
@@ -17,11 +17,12 @@
 	}
 
 	.navButtonContainer {
-		overflow: hidden;
-		z-index: 10;
-		top: ~"calc(50% - " (@sand-button-height + 12) / 2 ~")";
 		position: absolute;
 		width: 100%;
+		height: @sand-button-height * 2;
+		top: ~"calc(50% - " (@sand-button-height * 2) / 2 ~")";
```
MikyungKim, line 23: It'd be better if we make `@sand-button-height * 2` as a less variable for the future.

PageViews/PageViews.module.less
```diff
@@ -17,11 +17,12 @@
...
+		z-index: 10;
+		overflow: hidden;
```
MikyungKim, line 25: As we discussed, I think removing `overflow: hidden;` will solve the problem.

juwonjeong: It's fixed.

MikyungKim (approved):
LGTM

PR #1708, opened 2024-09-30
WRR-4660: Fixed `PageViews` to not clip the shadow of navigation buttons

juwonjeong:

### Checklist

The issue is that the navButtonContainer area of the PageViews was too tight, so the shadow of navigation buttons was clip when `fullContents` prop is `true`.

### Resolution
Set the height of the `navButtonContainer` to twice the height of the button.
And the formula related to `top`is a calculation that will position the navButtonContainer in the center.

### Additional Considerations

### Links
WRR-4660

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The following is a curated list of changes in the Enact sandstone module, newest
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


MikyungKim:
Let's update PR description too.

MikyungKim (approved):
LGTM

PR #1722, opened 2024-10-11
WRR-976: ContextualPopupDecorator: Modified to update position when screen orientation change

mmyelyn:
### Checklist

When ContextualPopupDecorator or ContextualMenuDecorator is open, rotating the device in the inspector should change the popup's position. Currently the popup is fixed and does not update.

### Resolution
I needed to check whether the position had not been updated or the value had been entered incorrectly. When rotated, I saw that when I refreshed, the position was calculated normally again. Therefore, the position calculation was no problem.

The position is updated with the positionContextualPopup function, which is not called when the window size changes, including when the screen is rotated. Therefore, I modified it to use ResizeObserver to detect changes in FloatLayer. 
(document.body may or may not have a size depending on the app)

### Additional Considerations

### Links
WRR-976

### Comments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: focal
+dist: jammy
 language: node_js
 node_js:
     - lts/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,7 @@
 dist: jammy
 language: node_js
 node_js:
-    - lts/*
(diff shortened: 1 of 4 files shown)


MikyungKim (approved):
LGTM

0x64 (approved):

PR #1726, opened 2024-10-18
WRR-6050: Fixed `IconItem` to restart marquee after done editing in `Scroller` with `editable` prop

MikyungKim:
## Checklist

When there is long `label` inside `IconItem` in editable `Scroller`, after done editing the marquee stops and it will never start again.

### Resolution
When completing edit, the DOM changes its order and it makes the marquee stops. I guess it's from engine but unfortunately, I coundn't find the specific logic or part. As far as I investigate, JS code has nothing to do with stopping maruee.
To resolve this issue, I chose to let `Marquee` to invalidate its metrics to restart safely with `invalidateProps` of `MarqueeDecorator`. I've added `order` as `invalidateProps` and pass it to `Marquee` through `IconItem` in JSX so that let `Marquee` know when the editing is done.
For the case where the editing ended without DOM change, I changed `Spotlight.focus` call added from https://github.com/enactjs/sandstone/pull/1550/files#diff-22d263c3907b8e109699fc727d876991ef576e0c70458dd2e6d59e014f7e426cR573-R574 conditionally to prevent stopping marquee.

### Additional Considerations
Currently there is no sample to test the problem. So you should change qa-sampler/IconItem.js to have `label` based on the change I made.

### Links
WRR-6050

### Comments

### DIFF
--- a/samples/sampler/stories/qa/IconItem.js
+++ b/samples/sampler/stories/qa/IconItem.js
@@ -64,7 +64,7 @@ const populateItems = ({index}) => {
 		} : null,
 		label: (function () {
 			if (index === 1) return 'USB';
-			else if (index === 6) return 'Gallery';
+			else if (index === 6) return 'Gallery label that has very long text';
 		})(),
 		labelColor: index === 6 ? 'dark' : null,
 		labelOn: index === 6 ? 'focus' : null,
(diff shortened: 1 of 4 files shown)


samples/sampler/stories/qa/IconItem.js
```diff
@@ -288,8 +288,10 @@ export const EditableIcon = (args) => {
 											className={css.editableIconItem}
 											css={css}
 											disabled={item.iconItemProps['disabled'] || item.hidden}
+											label={item.iconItemProps['label']}
```
0x64, line 291: I believe we don't need this. :)

0x64 (approved):

MikyungKim:
We will revisit the dropped coverage issue later.

seunghoh (approved):

PR #1740, opened 2024-11-12
WRR-9115: `Scroller`: Fixed to focus properly when the spottable node is bigger than the viewport by scrolling with voice control

mmyelyn:
### Checklist

When scrolling down with voice control in certain apps, focus is not restored.
Currently, when scrolling with voice control, focus is moved only when the bounds of the viewport include all bounds of the node.

Therefore, it is necessary to handle the case where the node's bounds are not all included in the viewport's bounds.
(when there is a large node that goes out of bounds, the upper part is visible but the lower part is cut off, etc.)

### Resolution
Added focus logic for cases where the top of the node is long and the bottom of the node is long using node Bounds and viewports Bounds. Since I need to receive and check new nodes, I implemented it using a recursive function.

### Additional Considerations

### Links
WRR-9115

### Comments


CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact sandstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `sandstone/Scroller` to focus even when the node boundery goes out the viewport by voice control
```
```suggestion
- `sandstone/Scroller` to focus properly when the spottable node is bigger than the size of viewport by voice control
```

useScroll/useEvent.js
```diff
@@ -449,15 +489,7 @@ const useEventVoice = (props, instances) => {
...
-				const nodes = Spotlight.getSpottableDescendants(scrollContainerNode.dataset.spotlightId);
-				for (let i = 0; i < nodes.length; i++) {
-					const nodeBounds = nodes[i].getBoundingClientRect();
-
-					if (nodeBounds[first] >= viewportBounds[first] && nodeBounds[last] <= viewportBounds[last]) {
-						Spotlight.focus(nodes[i]);
-						break;
-					}
-				}
+				focusSpottableNode(scrollContainerNode.dataset.spotlightId, viewportBounds, false);
```
MikyungKim, line 492: Let's use `getDeepSpottableDescendants` instead of this as we discussed.

mmyelyn: That's a concise answer. Thank you

MikyungKim (approved):
LGTM

juwonjeong (approved):
LGTM

PR #1750, opened 2024-11-15
WRR-6683: Remove logic to focus a slider when dragging with touch 

0x64:
### Checklist

When a user drags a slider with touch to move the slider, a console error occurs.
image

### Resolution
The console error occurs because a _ref_ is not properly passed to ui/Slider so I created the patch first. But, after investigation and tests, I concluded that the logic using the _ref_ is actually not needed anymore. Therefore, I want to remove the issued logic.

### Additional Considerations
The issued logic is to focus a slider knob even a user touch slightly out of the slider knob to show a focus effect on the knob. We don't need this in the current Sandstone.

### Links
WRR-6683

### Comments


daniel-stoian-lgp (approved):

juwonjeong (approved):
LGTM

PR #1780, opened 2024-12-12
WRR-9415: Replace .vendor-opacity mixin with the standard opacity attribute

0x64:
### Checklist

Some components are using enact/ui's .vendor-opacity() mixin but we don't need it. We normally use the standard opacity CSS attribute and just two components are using the mixin as their code is copied from the old codebase.

### Resolution
Replace .vendor-opacity() mixin with the standard opacity CSS attribute.

### Additional Considerations

### Links
WRR-9415

### Comments

### DIFF
--- a/IconItem/IconItem.module.less
+++ b/IconItem/IconItem.module.less
@@ -89,7 +89,7 @@
 		}
 
 		.disabled ({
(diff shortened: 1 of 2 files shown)


MikyungKim (approved):
LGTM

PR #1784, opened 2024-12-13
WRR-13724: Converted DebounceDecorator to functional component

ion-andrusciac-lgp:
### Checklist

Converted DebounceDecorator to functional component

### Resolution

### Additional Considerations

### Links
WRR-13724

### Comments

Enact-DCO-1.0-Signed-off-by: Ion Andrusciac ion.andrusciac@lgepartner.com

### DIFF
--- a/internal/DebounceDecorator/DebounceDecorator.js
+++ b/internal/DebounceDecorator/DebounceDecorator.js
@@ -73,7 +73,6 @@ const DebounceDecorator = hoc(defaultConfig, (config, Wrapped) => {
 
 		useEffect(() => {
 			return () => {
-				// eslint-disable-next-line react-hooks/exhaustive-deps
 				job.stop();
 			};
 			// eslint-disable-next-line react-hooks/exhaustive-deps
--- a/internal/DebounceDecorator/DebounceDecorator.js
+++ b/internal/DebounceDecorator/DebounceDecorator.js
@@ -8,7 +8,7 @@
 import hoc from '@enact/core/hoc';
 import {Job} from '@enact/core/util';
 import PropTypes from 'prop-types';
-import {Component} from 'react';
+import {useCallback, useEffect, useMemo} from 'react';
 
 /**
  * Default config for {@link sandstone/internal/DebounceDecorator.DebounceDecorator}.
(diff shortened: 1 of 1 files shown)


daniel-stoian-lgp (approved):
LGTM

internal/DebounceDecorator/DebounceDecorator.js
```diff
@@ -60,66 +60,66 @@ const defaultConfig = {
...
-
-		constructor (props) {
-			super(props);
-			this.job = new Job(this.emitEvent.bind(this), delay);
-		}
-
-		componentWillUnmount () {
-			this.job.stop();
-		}
+	const Debounce = (props) => {
```
0x64, line 63:
```suggestion
	// eslint-disable-next-line no-shadow
	const DebounceDecorator = (props) => {
```
We usually named a returned component as the same as the returning component for higher order components with eslint disabling rule for no-shadow. Not mandatory but if you don't mind... :)

internal/DebounceDecorator/DebounceDecorator.js
```diff
@@ -60,66 +60,66 @@ const defaultConfig = {
...
-		};
+		}, [props]);
 
-		handleCancel = (ev) => {
-			if (this.props[cancel]) {
-				this.props[cancel](ev);
-			}
-			this.job.stop();
-		};
+		const job = useMemo(() => new Job(emitEvent, delay), [emitEvent]);
```
0x64, line 72: I have a deep concern that what would be happened when `props` is changed in any reason during debounce logic works (=== meaning job.start() is called and job.stop() has not been called). In this case, `emitEvent` is re-created and `job` object is also re-created so that the old `job` is garbage collected without calling stop() function. It could make a call to an old `emitEvent` function object that is saved in a closure when `delay` is passed. What do you think about this?

0x64: Also, React does NOT guarentee that a cached object via `useMemo` is always kept when dependencies are not change. React will do it's best to keep the cached one but it does not guarantee. So, the returning value within `useMemo` should be pure: always same. But a job object returned from `new Job()` is not a pure object actually. It has a state by start and stop. We may need to find other safe option. Or, we can leave this code after we confirm that there is no risky case.

0x64:
Note:

- In this PR, we tries to convert a stateful HoC to a stateless HoC based on hooks so that the PR is resolving an issue of closure that occurs while we handle a state. Still, there is a small concern that the existing job is canceled when `props` is updated but we can handle it also if we really use this PR.
- For a reference purpose, I suggested to keep this PR even we don't merge this into our stable branch.

PR #1787, opened 2024-12-17
WRR-11058: Update editable scroller focus behavior to match the latest UX

juwonjeong:
### Checklist

Focus behavior of the editable scroller is updated
(1) After moving focus to the top button(item is not selected), focus will be moved by pressing the ← or → keys
241217_focusbehavior_1
(2) After entering the item edit mode, item can be moved by pressing the ← or → keys while focusing on the top button
241217_focusbehavior_2

### Resolution
Fix `handleKeyDownCapture` to move focus according to updated UX guide

### Additional Considerations
(1) To fix the issue as the gif below, I added 6084876fa63612a46d892ecf9162886ea6f91150 commit. 
241217_lastInputType
The issue occurs when focus moves to the `hide` button, scroll event occurred and then `handleMoveItemsByScroll` executed.
I added to set `lastInputType` to `key` to prevent `handleMoveItemsByScroll` from being executed.

(2) Since there is no guide for these cases, I implemented it as follows.
241217_edge_1
241217_edge_2
241217_edge_3
241217_edge_4

(3) I cannot use Spotlight.getLastInputType() in editableWrapper.js because it is later synced. so we had to define `lastInputType` in mutableRef. 

(4) I added 92fd7fcb63fee36d4dcfbfcae6bf2b74e00da38f, 73d5a95d261491e35ec7334f6692a8f8d78a7080 commit to fix the a11y issue which caused by the logic which manually focus the item. 

### Links
WRR-11058

### DIFF
--- a/Scroller/EditableWrapper.js
+++ b/Scroller/EditableWrapper.js
@@ -224,7 +224,7 @@ const EditableWrapper = (props) => {
 
 			setTimeout(() => {
 				if (item?.children[1]) {
-					item.children[1].ariaHidden = true;
+					item.children[1].ariaLabel = '';
 				}
 				if (!mutableRef.current.initialSelected) {
 					announceRef.current.announce(
@@ -582,9 +582,6 @@ const EditableWrapper = (props) => {
 		const {selectedItem, selectedItemLabel} = mutableRef.current;
 		const focusTarget = selectedItem.children[1];
 		const orders = finalizeOrders();
-
-		selectedItem.children[1].ariaLabel = '';
-		selectedItem.children[1].ariaHidden = false;
 
 		finalizeEditing(orders);
 		if (selectItemBy === 'press') {
--- a/Scroller/EditableWrapper.js
+++ b/Scroller/EditableWrapper.js
@@ -645,8 +645,6 @@ const EditableWrapper = (props) => {
 		const {focusedItem, selectedItem} = mutableRef.current;
 		const targetItemNode = findItemNode(target);
 
-		mutableRef.current.lastInputType = 'key';
-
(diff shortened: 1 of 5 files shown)


Scroller/EditableWrapper.js
```diff
@@ -623,45 +657,38 @@ const EditableWrapper = (props) => {
...
 
-				// Check if focus leaves scroll container.
-				if (nextTarget && !getContainersForNode(nextTarget).includes(mutableRef.current.spotlightId) && !repeat) {
-					reset();
+					setPointerMode(false);
+					Spotlight.focus(selectedItem.children[1]);
+				// If keyDown event target is button and next spot target is button, check whether focus leaves the scroll container.
+				} else {
+					// Set lastInputType to 'key' to prevent `handleMoveItemsByScroll` from being executed unexpectedly.
+					mutableRef.current.lastInputType = 'key';
```
0x64, line 675: I understand why we need this as it looks not easy to cover the case regarding global handlers and normal handlers. But, still I'm wonder whether we're able to gather the cases of setting `lastInputType` to 'key'. 😉 In general, the state cycle of `lastInputType` is not as same as our other code since it does not purely depend on the last input type but which key is pressed for key inputs.

SkylerBaek:
We still do not set 'key' to `lastInputType` no non-global keydown handler. Is this intended?

SkylerBaek:
> We still do not set 'key' to `lastInputType` no non-global keydown handler. Is this intended?

Sorry, I got it. Handlers are on capture phase. Not an issue.

SkylerBaek (approved):

MikyungKim (approved):
LGTM

PR #1805, opened 2025-01-21
WRR-17900: Fixed `PageViews` to focus spottable components on the same line with arrows for navigation when `fullContents`

MikyungKim:
### Checklist

When spottable components on the same line with arrows for navigation when `fullContents` of `PageViews`, the component cannot get the focus.

### Resolution
The reason why the issue is that PageView has an extra DOM for aligning arrows for navigation above the content and it blocks the spottable components to focus.
I added `pointer-events: none` for the DOM that doesn't need to consume the events.

### Additional Considerations

### Links
WRR-17900

### Comments


juwonjeong (approved):
LGTM

PR #1806, opened 2025-01-23
WRR-17701: Fix editable scroller to move focus by holding directional key

juwonjeong:

### Checklist

* [x] I have read and understand the [contribution guide](https://enactjs.com/docs/developer-guide/contributing/)
* [x] A [CHANGELOG entry](https://enactjs.com/docs/developer-guide/contributing/changelogs/) is included

When pressing(holding) directional key(e.g., up key) in editable scroller, focus needs to move continusly

### Resolution
As I analyzed, focus failed to move in `handleFocusLeaveScrollContainer` because focus is prevented if 5-way is being held and the next element isn't wrapped by the current element's container. 
related code: https://github.com/enactjs/enact/blob/master/packages/spotlight/src/spotlight.js#L354
So I changed the Spotlight.move() function to Spotlight.focus(nextTarget) to avoid this issue. 

### Additional Considerations

### Links
WRR-17701

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The following is a curated list of changes in the Enact sandstone module, newest
 
 - `sandstone/TooltipDecorator` to hide a tooltip when tapping outside of disabled component
 - `sandstone/Scroller` with `editable` prop focus behavior to match the latest UX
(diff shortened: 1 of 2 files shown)


MikyungKim (approved):
LGTM

PR #1811, opened 2025-02-10
WRR-19431: Fixed WizardPanels so that content does not move when moving to the next panel by adding panels dynamically

mmyelyn:
### Checklist

* [X] I have read and understand the [contribution guide](https://enactjs.com/docs/developer-guide/contributing/)
* [X] A [CHANGELOG entry](https://enactjs.com/docs/developer-guide/contributing/changelogs/) is included

When you move to the next panel in WizardPanels, the content position of the first panel changes briefly. 
This occurs when the App dynamically adds a Panel to WizardPanels.

### Resolution
When a panel is added dynamically, a space has been added to set the height of the title when moving to the next panel.

### Additional Considerations
We need to think about the subtitles' height and Heading's line-height.

### Links
WRR-19431

### Comments

### DIFF
--- a/Panels/Header.js
+++ b/Panels/Header.js
@@ -371,7 +371,6 @@ const HeaderBase = kind({
 					alignment={centered ? 'center' : null}
 					className={css.title}
 				>
-					{/* WRR-19431: When a panel is added dynamically, a space has been added to set the height of the title when moving to the next panel.*/}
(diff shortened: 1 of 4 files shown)


Panels/Header.js
```diff
@@ -438,7 +440,6 @@ const HeaderBase = kind({
 		slotBeforeRef,
 		slotSize,
 		titleCell,
-		type,
```
MikyungKim, line 441: Let's delete `type` from `rest`.

MikyungKim (approved):
LGTM

